PR #1, opened 2021-08-20
add: kube-scheduler-simulator

sanposhiho:
This PR adds all codes of simulator🎉 

/cc @adtac


Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+WORKDIR /go/src/simulator-server
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+RUN go build -v -o ./bin/simulator simulator.go
+
+FROM alpine:3.14.0
```
adtac, line 16: +1 for separating build and final image :+1:

server/di/di.go
```diff
@@ -0,0 +1,68 @@
+package di
```
adtac, line 1: nit: what does this stand for? can we use a more descriptive name for this package?

Issue #8, opened 2021-08-30
Make it easy for users to add their custom plugins


sanposhiho:
Currently, the simulator only supports default plugins. 
But some users may want to try their custom plugins on the simulator.

- Describe in the documentation how to add a custom plugin to the simulator.
- Add some functions(or some option?) to register the custom plugins

/assign

warmchang:
@sanposhiho thank you for this awesome project!

Here is a blank should be trim ('.dockerignore  '), otherwise the git clone don't work on windows.
(a small typo, so no need to open a separate issue to track. :smile: )
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/.dockerignore%20%20

```
william@WIN-MNJRNPMAROQ MINGW64 /d/thehub
$ git clone https://github.com/kubernetes-sigs/kube-scheduler-simulator
Cloning into 'kube-scheduler-simulator'...
remote: Enumerating objects: 209, done.
remote: Counting objects: 100% (209/209), done.
remote: Compressing objects: 100% (155/155), done.
remote: Total 209 (delta 36), reused 196 (delta 33), pack-reused 0
Receiving objects: 100% (209/209), 1.19 MiB | 181.00 KiB/s, done.
Resolving deltas: 100% (36/36), done.
error: invalid path '.dockerignore  '
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'


william@WIN-MNJRNPMAROQ MINGW64 /d/thehub
$
```

sanposhiho:
Hi @warmchang, thanks for reporting. 
I'll fix it soon 🤦‍♂️

sanposhiho:
Aha, sorry. You said about custom scheduler, not custom plugins. 
I think if your custom scheduler has the same interface as the original scheduler, you can try your custom scheduler with the simulator by changing all internal import statements in the simulator to use your scheduler instead of the original scheduler.

sanposhiho:
/close

Close this issue. This issue will be resolved with #35

sanposhiho:
> This issue will be resolved with https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/35

Oh, really? I don't think so. 😅 

/reopen

PR #9, opened 2021-08-30
Fix: file name .dockerignore has blank

sanposhiho:
Hello team.
.dockerignore has blank on its filename and git cannot checkout successfully on windows.

Thanks for reporting, @warmchang.

https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/8#issuecomment-908180585


adtac:
The file is still empty? Did you mean to add a newline? I don't see it in the diff

sanposhiho:
@adtac 
It looks like there is no change, but actually the file name has been changed :)

`.dockerignore  `  (two whitespaces after name) → `.dockerignore` (delete two whitespaces)

(You can tell from the URL that it contains two whitespaces (`%20%20`).)
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/.dockerignore%20%20

sanposhiho:
@adtac ping

Issue #12, opened 2021-08-30
failed to restart scheduler if specify .profiles.pluginConfig 


sanposhiho:
## how to reproduce

restart scheduler with specifying .profiles.pluginConfig 

## log
```
E0830 22:49:48.480481   32087 schedulerconfig.go:36] failed to restart scheduler: start scheduler:
    github.com/kubernetes-sigs/kube-scheduler-simulator/scheduler.(*Service).RestartScheduler
        /Users/kenseinakada/workspace/kube-scheduler-simulator/scheduler/scheduler.go:42
  - create scheduler:
    github.com/kubernetes-sigs/kube-scheduler-simulator/scheduler.(*Service).StartScheduler
        /Users/kenseinakada/workspace/kube-scheduler-simulator/scheduler/scheduler.go:90
  - couldn't create scheduler: initializing profiles: creating profile for scheduler name default-scheduler: initializing plugin "NodeAffinityForSimulator": create original plugin: args are not of type NodeAffinityArgs, got *runtime.Unknown
```

/kind bug
/assign

PR #13, opened 2021-08-30
override configuration correctly, even if passed PluginConfig has args data on PluginConfig.Args.Raw

sanposhiho:
Hi team~

Fixes: #12 

This PR fixes the bug on restarting scheduler. 
The function NewPluginConfig failed to override configuration correctly when passed PluginConfig has args data on PluginConfig.Args.Raw not on PluginConfig.Args.Object. So, fixed it

/kind bug

### DIFF
--- a/scheduler/plugin/plugins.go
+++ b/scheduler/plugin/plugins.go
@@ -2,6 +2,7 @@ package plugin
 
 import (
 	"context"
+	"encoding/json"
 
 	"golang.org/x/xerrors"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


scheduler/plugin/plugins.go
```diff
@@ -70,30 +71,49 @@ func NewRegistry(informerFactory informers.SharedInformerFactory, client clients
...
+		pluginConfig[name] = &defaultcfg.Profiles[0].PluginConfig[i].Args
 	}
 
 	for i := range pc {
 		name := pc[i].Name
-		// override default configuration
-		pluginConfig[name] = pc[i].Args
+		ret := pluginConfig[name].DeepCopy()
+
+		// v1beta2.PluginConfig may have data on pc[i].Args.Raw as []byte.
```
haosdent, line 94: have data on -> have data in

scheduler/plugin/plugins.go
```diff
@@ -70,30 +71,49 @@ func NewRegistry(informerFactory informers.SharedInformerFactory, client clients
...
+		// We have to encoding it in this case.
+		if len(pc[i].Args.Raw) != 0 {
+			// override default configuration
+			if err := json.Unmarshal(pc[i].Args.Raw, &ret.Object); err != nil {
+				return nil, err
+			}
+		}
+
+		if pc[i].Args.Object != nil {
+			// If Args data exists in both PluginConfig.Args.Raw and PluginConfig.Args.Object, PluginConfig.Args.Object takes precedence.
```
haosdent, line 104: If Args data -> If data

haosdent: PluginConfig.Args.Object takes precedence -> PluginConfig.Args.Raw would be ignored

scheduler/plugin/plugins_test.go
```diff
@@ -257,6 +258,63 @@ func Test_NewPluginConfig(t *testing.T) {
...
+									HardPodAffinityWeight: &hardPodAffinityWeight,
+								},
+							},
+						}
+					}
+				}
+
+				return pc
+			}(),
+			wantErr: false,
```
haosdent, line 316: Would you like to add a test case to verify
> 			// If Args data exists in both PluginConfig.Args.Raw and PluginConfig.Args.Object, PluginConfig.Args.Object takes precedence.

scheduler/plugin/plugins.go
```diff
@@ -70,30 +71,49 @@ func NewRegistry(informerFactory informers.SharedInformerFactory, client clients
 
 // NewPluginConfig converts []v1beta2.PluginConfig for simulator.
 // Passed []v1beta.PluginConfig overrides default config values.
+//
+// NewPluginConfig expects that passed v1beta2.PluginConfig has data of Args in only one of PluginConfig.Args.Raw or PluginConfig.Args.Object.
```
haosdent, line 75:
```
NewPluginConfig expects that either PluginConfig.Args.Raw or PluginConfig.Args.Object has data in the passed v1beta2.PluginConfig parameter.
If data exists in both PluginConfig.Args.Raw and PluginConfig.Args.Object, PluginConfig.Args.Raw would be ignored since PluginConfig.Args.Object has higher priority.
```

haosdent (commented):
I rephrase some comments, but I am not a native speaker, can ignore them if you think the comment is incorrect.

scheduler/plugin/plugins_test.go
sanposhiho: Sure. 👍

sanposhiho:
@haosdent 
Thanks for reviewing. 
I fixed the wording of comments as your suggestions and added a test case.

PR #15, opened 2021-08-31
Enable to check if etcd is on path in start_simulator.sh

sanposhiho:
Hello team.

This PR enable to check if etcd is on path in start_simulator.sh.

### DIFF
--- a/hack/start_simulator.sh
+++ b/hack/start_simulator.sh
@@ -2,7 +2,7 @@
 
 source "./hack/etcd.sh"
 
-checkEtcdOnPath() {
+check_etcd_on_path() {
   echo "Checking etcd is on PATH"
   which etcd && return
   echo "Cannot find etcd on PATH."
@@ -29,8 +29,7 @@ cleanup_etcd() {
   echo "Clean up finished"
 }
 
-
-checkEtcdOnPath
+check_etcd_on_path
 
 start_etcd
 trap cleanup_etcd EXIT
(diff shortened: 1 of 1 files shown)


hack/start_simulator.sh
```diff
@@ -29,8 +29,10 @@ cleanup_etcd() {
   echo "Clean up finished"
 }
 
-trap cleanup_etcd EXIT
 
+checkEtcdOnPath
+
+trap cleanup_etcd EXIT
```
adtac, line 35: trap after start_etcd please

sanposhiho: fixed thanks

hack/start_simulator.sh
```diff
@@ -29,9 +29,11 @@ cleanup_etcd() {
   echo "Clean up finished"
 }
 
-trap cleanup_etcd EXIT
+
+checkEtcdOnPath
```
haosdent, line 33: This fun name looks weird. How about use checkIfEtcdExists

hack/start_simulator.sh
```diff
@@ -8,7 +8,7 @@ checkEtcdOnPath() {
   echo "Cannot find etcd on PATH."
```
haosdent, line 8: PATH -> $PATH

hack/start_simulator.sh
haosdent: Better to use check_if_etcd_exists to avoid we mix snake style and camel style in this file.

sanposhiho:
@haosdent 
Thanks for reviewing. I fixed some wording as your suggestions.

haosdent:
you are welcome @sanposhiho

196Ikuchil:
I ran `./hack/start_simulator.sh` in my local, and verified that the messages are printed if `which etcd` command fails (etcd is not installed).

/lgtm

PR #18, opened 2021-08-31
Workaround: v-data-table cannot scroll when the bar cannot be scrolled vertically

### LINKED ISSUE #10: cannot see all of table and cannot even scroll when the browser width is small
like this. In this example, we cannot see the result of some plugins because the width is not enough.

スクリーンショット 2021-08-30 22 32 42

/kind bug
/assign

sanposhiho:
Fixes #10 
It seems that v-data-table(vuetify component used to show table) cannot scroll horizontally when the v-navigation-drawer cannot be scrolled vertically 🤔  and it causes bug #10.

I add `<div>` with height as hot fix so that v-navigation-drawer always can be scrolled vertically.

/kind bug

### DIFF
--- a/web/components/ResourceBar/ResourceBar.vue
+++ b/web/components/ResourceBar/ResourceBar.vue
@@ -32,6 +32,8 @@
     <template v-if="!editmode">
       <SchedulingResults />
       <ResourceDefinitionTree :items="treeData" />
+      <!-- This is need to avoid bug on vuetify -->
+      <div style="height: 80%;"></div>
     </template>
   </v-navigation-drawer>
 </template>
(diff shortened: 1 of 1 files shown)


web/components/ResourceBar/ResourceBar.vue
```diff
@@ -32,6 +32,8 @@
     <template v-if="!editmode">
       <SchedulingResults />
       <ResourceDefinitionTree :items="treeData" />
+      <!-- This is need to avoid bug on vuetify -->
```
haosdent, line 35: This is required to work around the vuetify's bug, refer more details in https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/10

sanposhiho: @haosdent 
Thanks for the suggestion. 
Fixed. 🙏

sanposhiho:
@haosdent 
Could you `/lgtm` if looks good to you? Or expect another reviewer to review it?

Issue #19, opened 2021-09-02
Provide images for simulator with k8s.gcr.io


sanposhiho:
/kind feature
/cc @adtac 
/cc @Huang-Wei 
/cc @alculquicondor 
/assign

Hi team. 

## what want to do

add images for backend/frontend to k8s.gcr.io
- create PR on kubernetes/k8s.io 
  - this doc helps us: https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io
  - (like this pr?: https://github.com/kubernetes/k8s.io/pull/1431/files)
- push images with cloudbuild 
  - this doc helps us: https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md

## why

Now, when users want to use the simulator, they have to clone this repo, build it and run it(on docker or local). 
This is a little inconvenient because it will take them a lot of time to build frontend/backend (maybe because of some huge dependencies).

So it is great if we can provide images for backend/frontend so that users can only have to run these images. wdyt?

## Question

- Are there any requirements for creating the image?
- Do we need to get someone's approval to create images? 

I cannot find anything about these in documentations.

alculquicondor:
What other kind of requirements are you referring to?

I would image that the only people that need to approve are the OWNERS of kubernetes/k8s.io and kubernetes/test-infra

sanposhiho:
Thanks, so I can get theirs approves when I create the PRs.
I was just wondering if there are any special requirements since we will be using GCR resources. If not, it's OK. 👍 

And we have to create a google group for push access. Who should we list?
> Create a google group for granting push access by adding an email alias for it in groups.yaml. 
https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io#creating-staging-repos

alculquicondor:
Probably the same as the owners

sanposhiho:
Okay, I'll get your email addresses from the email list of sig/scheduling.

Huang-Wei:
@sanposhiho To promote an image to gcr.io, it's two-phased. First the CI job usually builds images on every commit / tag, to staging.gcr.io, and secondly, the project maintainer manually promotes some particular images as the public available ones to gcr.io.

[...]

PR #21, opened 2021-09-12
Support build with cloudbuild for gcr

sanposhiho:
part of #19 

guide: https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md

For the most part, it is almost the same as the example in the guide.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,32 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+
+# this must be specified in seconds. If omitted, defaults to 600s (10 mins)
+timeout: 1200s
+steps:
+  - name: gcr.io/cloud-builders/docker
+    args:
+    - build
+    - --tag=gcr.io/$PROJECT_ID/simulator-backend:$_GIT_TAG
+    - --tag=gcr.io/$PROJECT_ID/simulator-backend:latest
+    - -f=Dockerfile
+    - .
+  - name: gcr.io/cloud-builders/docker
+    args:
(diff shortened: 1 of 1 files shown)


sanposhiho:
image pushing job will be created in https://github.com/kubernetes/test-infra/pull/23567

cloudbuild.yaml
```diff
@@ -0,0 +1,32 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+
+# this must be specified in seconds. If omitted, defaults to 600s (10 mins)
+timeout: 1200s
+steps:
+  - name: gcr.io/cloud-builders/docker
```
Huang-Wei, line 6: Is this conventional? I noticed some repos used `gcr.io/k8s-staging-test-infra/gcb-docker-gcloud`.

sanposhiho: > If your build process is driven by a Makefile or similar, you can use GCB to invoke that. We provide the gcr.io/k8s-testimages/gcb-docker-gcloud image, which contains components that are likely to be useful for your builds. 
https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md#makefile-build-example

So, `gcr.io/k8s-staging-test-infra/gcb-docker-gcloud` image is for Makefile build, maybe...? 

Because simple build example uses `gcr.io/cloud-builders/docker` (and I followed it.)
https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md#simple-build-example

PR #30, opened 2021-09-30
Update README.md about memory allocation to Docker

### LINKED ISSUE #29: `make docker_build_and_up` failed
An error occurred when starting up the Kubernetes scheduler simulator using Docker.

### environment

- M1 MacBook Pro
- Docker for Mac (darwin/arm64)
- docker engine 20.10.8

### Command

```
make docker_build_and_up
```

### Error Message

```
executor failed running [/bin/sh -c yarn build && yarn install --production --ignore-scripts --prefer-offline]: exit code: 1
make: *** [docker_build_front] Error 1
```

### Comment

/kind bug

riita10069:
Fixes #29

/kind bug

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,8 @@ make docker_build_and_up
 
 Then, you can access the simulator with http://localhost:3000
 
-Note: 4GB of memory should be allocated to Docker.
-If you do not allocate enough memory, it may fail.
+Note: Insufficient memory allocation sometimes causes problems in building the image.
+Please allocate enough memory in that case.
 
 ### Run Locally
 
--- a/README.md
+++ b/README.md
@@ -53,6 +53,9 @@ make docker_build_and_up
 
 Then, you can access the simulator with http://localhost:3000
 
+Note: 4GB of memory should be allocated to Docker.
+If you do not allocate enough memory, it may fail.
+
 ### Run Locally
 
 You have to run frontend, server and etcd.


README.md
```diff
@@ -53,6 +53,9 @@ make docker_build_and_up
 
 Then, you can access the simulator with http://localhost:3000
 
+Note: 4GB of memory should be allocated to Docker.
```
sanposhiho, line 56: We haven't exactly figured out the cause. This problem may occur only in docker for mac. (I mean it may not always require 4GB momory)
So, lMO it is better to change the wording to something a little more ambiguous.

like: 
```
Note: Insufficient memory allocation sometimes causes problems in building the image.
Please allocate enough memory in that case.
```

sanposhiho (commented):
thanks

sanposhiho (commented):
@riita10069 can you change this PR description to like `Fixes #29 `

Because GitHub automatically close the issue after this PR is merged
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

README.md
alculquicondor: 4GB sounds like a lot. What is consuming that? the frontend or the backend?

riita10069: Failing to build an Image.

I had a problem with 2GB memory allocation.
With 4GB, it did not happen.

sanposhiho (approved):
I've been investigating for a while, but I couldn't find any fundamental solution to this issue. 

It seems that there are cases where something uses a lot of memory during the build process of nodejs. In some cases, such problems can be solved by introducing `NODE_OPTIONS="–-max_old_space_size=XXX"`, but it did not solve our problem.

So, I'd like to accept this change -- The simplest solution is to increase the memory allocation.
And, when we provide the image for the simulator (#19), users will not have to build the image themselves, so this will be less of a problem.

/lgtm

PR #32, opened 2021-10-11
Added restore old setting function

### LINKED ISSUE #11: restore the original settings and reboot scheduler when fail to restart scheduler with user-custom settings
## Background

this simulator restart when user request to change the scheduler configuration.
Currently, when fail to restart with custom setting, the scheduler is dead until the user adapts the correct settings.

## new feature

restore the original settings and reboot scheduler when fail to restart scheduler with user-custom settings

tamboliasir1:
Fix #11


tamboliasir1:
Got it.. I'll just get the old settings and restart it with restart scheduler..

Thank you for guidance

PR #36, opened 2021-10-30
Updated restart scheduler for restore old setting

### LINKED ISSUE #11: restore the original settings and reboot scheduler when fail to restart scheduler with user-custom settings

tamboliasir1:
Thank you for your support and guidance.
Fix #11

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -41,7 +41,10 @@ func (s *Service) RestartScheduler(cfg *v1beta2config.KubeSchedulerConfiguration
 	oldSchedulerCfg := s.currentSchedulerCfg
 	if err := s.StartScheduler(cfg); err != nil {
 		klog.Info("Applying old configuration setting")
-		s.StartScheduler(oldSchedulerCfg)
+		if err := s.StartScheduler(oldSchedulerCfg); err != nil {
+			klog.Warningf("failed to start simulator server: %v", err)
+			return xerrors.Errorf("start scheduler: %w", err)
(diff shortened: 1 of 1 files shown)


scheduler/scheduler.go
```diff
@@ -38,7 +38,10 @@ func NewSchedulerService(client clientset.Interface, restclientCfg *restclient.C
 func (s *Service) RestartScheduler(cfg *v1beta2config.KubeSchedulerConfiguration) error {
 	s.ShutdownScheduler()
 
+	oldSchedulerCfg := s.currentSchedulerCfg
 	if err := s.StartScheduler(cfg); err != nil {
+		klog.Info("Applying old configuration setting")
+		s.StartScheduler(oldSchedulerCfg)
```
sanposhiho, line 44: I think it is better to check the error from this second `s.StartScheduler` and the error is printed as warn log when `err != nil` 🙏

sanposhiho (commented):
Thanks :)

sanposhiho (approved):
Nice. Thinks for your contribution :)
/lgtm

sanposhiho (commented):
Only nits.

sanposhiho (approved):
Thanks :)
/lgtm
/approve

Issue #38, opened 2021-11-04
cannot build Docker frontend due to error from nodejs


sanposhiho:
This issue is due to https://github.com/webpack/webpack/issues/14532

```
#15 5.142 Error: error:0308010C:digital envelope routines::unsupported
#15 5.142     at new Hash (node:internal/crypto/hash:67:19)
#15 5.142     at Object.createHash (node:crypto:130:10)
#15 5.142     at module.exports (/app/node_modules/webpack/lib/util/createHash.js:135:53)
#15 5.142     at NormalModule._initBuildHash (/app/node_modules/webpack/lib/NormalModule.js:417:16)
#15 5.142     at handleParseError (/app/node_modules/webpack/lib/NormalModule.js:471:10)
#15 5.142     at /app/node_modules/webpack/lib/NormalModule.js:503:5
#15 5.142     at /app/node_modules/webpack/lib/NormalModule.js:358:12
#15 5.142     at /app/node_modules/loader-runner/lib/LoaderRunner.js:373:3
#15 5.142     at iterateNormalLoaders (/app/node_modules/loader-runner/lib/LoaderRunner.js:214:10)
#15 5.142     at Array.<anonymous> (/app/node_modules/loader-runner/lib/LoaderRunner.js:205:4)
#15 5.142     at Storage.finished (/app/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:55:16)
#15 5.142     at /app/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:91:9
#15 5.142     at /app/node_modules/graceful-fs/graceful-fs.js:123:16
#15 5.142     at FSReqCallback.readFileAfterClose [as oncomplete] (node:internal/fs/read_file_context:68:3) {
#15 5.142   opensslErrorStack: [ 'error:03000086:digital envelope routines::initialization error' ],
#15 5.142   library: 'digital envelope routines',
#15 5.142   reason: 'unsupported',
#15 5.142   code: 'ERR_OSSL_EVP_UNSUPPORTED'
#15 5.142 }
```

## How to solve

I have confirmed that the solution described in [the issue](https://github.com/webpack/webpack/issues/14532) will solve the problem.

sanposhiho:
Adding `ENV NODE_OPTIONS --openssl-legacy-provider` like this will solve this issue.

[...]

PR #45, opened 2021-11-30
pin node image to Node.js v16

tk3fftk:
current latest node image is Node.js version v17. Using latest image should break codes unexpectedly.
Also, odd-numbered releases of Node.js is [short term version](https://nodejs.org/en/about/releases/).

Related: https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/38


sanposhiho (approved):
I also confirmed v16 works well.
Thanks :)

/approve
/lgtm

Issue #55, opened 2021-12-13
make internal kube-apiserver exposable and accessible


riita10069:
It is very difficult to recreate a cluster in production env, using the web.
To make it easier, I want to use `kubectl apply` command.

sanposhiho:
+1
I like this idea.
Note that we need to change docker-compose.yml to expose kube-apiserver in simulator backend. I think it is better to create new docker-compose.yml (or split existing docker-compose.yml) for this purpose, because this feature doesn't require web ui.

/retitle request to internal kube-apiserver with kubectl cmd

riita10069:
Thanks for the heads up!

sanposhiho:
And it would be nice to provide some way to see the result of plugins. 
It is difficult to see the result in pod annotation.

riita10069:
I can't access kubeapiserver from the local env.
I think the reason is that local env don't have the token needed for kubeapiserver access.
Do you have something good solutions?

```
Error from server (InternalError): an error on the server ("") has prevented the request from succeeding
```

It is possible to authenticate with Username and Password, and I am planning to investigate that.

sanposhiho:
Let me retitle again to make the purpose of this issue more general.
If we can expose internal kube-apiserver, it means we can access it in any way (like client-go)

/retitle make internal kube-apiserver exposable and accessible

sanposhiho:
By #94, we can achieve this and the rest work is only documentation, I think. 
The documentation should explain how to change the URL of kube-apiserver in the simulator and how to connect for each client (kubectl and client-go).

sanposhiho:
We can close this because https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/138 will add the doc for each env.
/close

Issue #27, opened 2021-09-27
support PriorityClass


sanposhiho:
Now we don't support PriorityClass.

If you don't know about PriorityClass and preemption, see this doc: https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/

We cannot simulate preemption without PriorityClass. So I'd like to add the feature to create/edit/delete PriorityClass.
The implementation may take some time because development of this feature requires changes both frontend and backend. Seeing implementation of other resources will help.

Please let me know if you have any questions.

/kind feature

khalilswdp:
Hi @sanposhiho, I am interested in contributing to this issue. Can you please assign this for me. Though this is my first open source contribution and I don't know how to do it. I am reading the documentation related to the issue, can you provide me with more resources to follow as a first timer?

khalilswdp:
I'd also appreciate if I get some insights on how to start the issue

sanposhiho:
Hi @khalilswdp 

The background of this issue can be found in this issue description. Obviously, we need to support PriorityClass, but not yet.
If you want to try this issue, I suggest you start with the backend.

In this dir, you can see how to add new handler for new resources in API server.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/tree/master/server/handler

Handlers are initialized and registered in `NewSimulatorServer`.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/server/server.go

And, for example, in this dir, you can see how to implement service layer. 
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/storageclass/storageclass.go

Interfaces for service layer should be included in this file.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/server/di/interface.go

And initializing service layer should be done in `NewDIContainer`.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/server/di/di.go

sanposhiho:
/reopen

@khalilswdp 
Could you update README like this?:
https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/69

khalilswdp:
@sanposhiho ok I'm on it

sanposhiho:
Finished supporting. Thanks @khalilswdp.
/close

khalilswdp:
Thank you @sanposhiho  for all the help, I am really grateful!

Issue #26, opened 2021-09-23
Add reset button to reset all resources and scheduler configuration


sanposhiho:
Add reset button

- delete all resources
- change scheduler configuration to default one.

 need to change both frontend and backend.

/kind feature

196Ikuchil:
Hi, @sivchari 
I'm came from #67 
So is this what you're saying?
We need to implement one service that has a function(method) to receive a Reset request and call "DeleteAll" method of each service. Or, implement it on a simulator service that already exists.

As one additional solution, I would like to propose that integrating with this #64 functions. These are similar to yours in terms of the scope of service its uses and manages. 
(and just as time, I'm having trouble deciding on this package name. )
what do you think?
Cc @sanposhiho

sanposhiho:
As I said before, we had two ways for this feature.

1. create one API to reset all resources and scheduler config.
2. create multiple APIs to reset each resource and scheduler config.

And @sivchari has already created new API to reset scheduler in #54. 
This means we take the way 2.

And, #68 creates "delete all" method in each service. So, in backend, all that remain tasks for us to do is to improve handler layers for each resource. 

@196Ikuchil 
> As one additional solution, I would like to propose that integrating with this #64 functions.

I cannot catch what your "integrating" means. 
At least, I don't think it is a good idea to implement this reset feature in the same package as import/export feature package in #64. If we do so, that package will be too big.

196Ikuchil:
Oops, it's "in each service". I was mistaken and off the main purpose, sorry...

@sanposhiho 
> I cannot catch what your "integrating" means.
This is means I just thought they were similar in that it calls each service if we had taken the way 1.

sanposhiho:
Yes, if we had taken the way 1, we would have created a new service to delete all resources.

sivchari:
Sorry, I was not clear enough.

> We need to implement one service that has a function(method) to receive a Reset request and call "DeleteAll" method of each service.

I was thinking of implementing something like this.
This is because if we want to call each handler, we need to call each request one at a time.
I mentioned this question because I thought this was not so good and it would be better to create a new handler and call DeleteAll for each service in one request.

/cc @sanposhiho 
/cc @196Ikuchil

sanposhiho:
@sivchari 
Got it. So, do you want to delete API to reset scheduler configuration created in #54?

sivchari:
I think #54 is API that only restart scheduler. This issue scopes to delete each service.
[...]

PR #64, opened 2021-12-19
Implement import/export API

### LINKED ISSUE #28: Add feature to export/import current resources
It would be nice if we can export the resources at that time and can import and restore the resources.

196Ikuchil:
#### What type of PR is this?
/kind api-change
/kind feature

- Add import/export method on API
  - Add resources service & handler

`export`
1. Call each `List` function and return it in `json`.

`import`
1. Post data in `json`.
2. Apply` to each resource.
3. return status 200 code if cuccess.
 

Fixes #28

Please review the sections I commented on.

/label do-not-merge/work-in-progress
/hold


resources/resources.go
```diff
@@ -0,0 +1,194 @@
...
+		_, err := s.pvcService.Apply(ctx, &pvc)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to apply PersistentVolumeClaims: %+v", err)
+		}
+	}
+	for _, pv := range *resources.PvList {
+		pv.ObjectMetaApplyConfiguration.UID = nil
+
+		// Status is omitempty. Is this field become nil?
+		if *pv.Status.Phase == "Bound" {
```
196Ikuchil, line 160: PersistentVolume is associated with PersistentVolumeClaims. Here make the association manually. Does a better way exist?

server/handler/resources.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (h *ResourcesHandler) ExportResourcesAll(c echo.Context) error {
+	ctx := c.Request().Context()
+
+	rs, err := h.service.ExportAll(ctx)
+	if err != nil {
+		klog.Errorf("failed to export all resources: %+v", err)
+		return echo.NewHTTPError(http.StatusInternalServerError)
+	}
+	return c.JSON(http.StatusOK, rs)
```
196Ikuchil, line 29: I'm going to convert to yaml format on the frontend.

server/handler/resources.go
```diff
@@ -0,0 +1,69 @@
...
+			klog.Errorf("failed to convert to ResourcesApplyConfiguration of recovery resources: %+v", err2)
+			return echo.NewHTTPError(http.StatusInternalServerError)
+		}
+		_, err3 := h.service.ImportAll(ctx, bkpResources)
+		if err3 != nil {
+			klog.Errorf("failed to recover of backup resources: %+v", err3)
+			return echo.NewHTTPError(http.StatusInternalServerError)
+		}
+	}
+	return c.JSON(http.StatusOK, rs)
```
196Ikuchil, line 68: Return result resources if succeed. 
I think that it is also good for call get api again from the client-side.

server/handler/resources.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	reqResources := new(resources.ResourcesApplyConfiguration)
+	if err := c.Bind(reqResources); err != nil {
+		klog.Errorf("failed to bind import resources all request: %+v", err)
+		return echo.NewHTTPError(http.StatusInternalServerError)
+	}
+
+	rs, err := h.service.ImportAll(ctx, reqResources)
+	if err != nil {
```
196Ikuchil, line 49: Here we restore the previous resource in case of failure.

196Ikuchil:
comment for refresh CLA

riita10069:
@196Ikuchil 
Thanks for your first contribution!!


I've been trying to solve this issue with this.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/55

I apologize for the delay in mentioning this issue.
May I have a discussion on it.

I'd like to discuss this point, because I think `kubectl get -o yaml` is more generic and usable.

This is just my opinion. There is room for discussion.


/cc @sanposhiho

196Ikuchil:
@riita10069 
Thanks for your comment! 
I hadn't noticed PR: #55 until now.

> I'd like to discuss this point, because I think kubectl get -o yaml is more generic and usable.
Yes, I think that idea is better, too. Because this my API requires manual rewriting of UIDs, etc. (or I just don't know a better way.)
So, `kubectl get -o yaml` is more simple and smart (and generic).


I'd like to know your opinion, so let me ask you a question.
The currently implemented Services (e.g. Node's Get, List, Apply..) will be also replaced with a way to use `kubectl`, if possible?

sanposhiho:
> I'd like to discuss this point, because I think kubectl get -o yaml is more generic and usable.

Hm, IMO, "we can do this with kubectl, client-go or other client" doesn't mean "we don't need to implement this in web UI".

kubectl is one of the clients of simulator, and web UI is also one of the clients of simulator.

To use kubectl for the simulator will need to configure it. Even if it is easy, it's a little hassle for users using the simulator via Web UI to do that.

Of course, we don't need to support "anything we can do with kubectl" in web UI. But, I think we need to support features that users are likely to use in the simulator.

riita10069:
@196Ikuchil 

Thanks for your replies.

> The currently implemented Services (e.g. Node's Get, List, Apply..) will be also replaced with a way to use kubectl, if possible?

Yes, I think we can.

> requires manual rewriting of UIDs

If we're going to implement it in client-go, I'd like to think about a better way.

@sanposhiho 

Since kubectl runs as a single binary, it can be easily run inside a container.
You can also use kubectl as an internal implementation for the same kind of experience. That way, we don't have to maintain support for Kubernetes frequent API updates.
Instead of that, all we have to do is just upgrade the version of kubectl.

This is also relevant to #55

196Ikuchil:
> You can also use kubectl as an internal implementation for the same kind of experience. That way, we don't have to maintain support for Kubernetes frequent API updates.

I think this is a big advantage. So, if almost all functions can be replaced by kubectl, I think the idea is good. 
If not, it will be a bit complicated to implement and maintain?

Also, I may be not agreed with using kubectl if there are parameters that cannot be obtained with kubectl, or if it goes against the purpose of this project.
(I'm not picky about it, so this opinion is just for reference purposes✋)

sanposhiho:
I talked with @riita10069 in a video call. 
We, now, use client-go to send requests to API-server, and the proposal from @riita10069 is replacing it with kubectl.
But, depending on external command-line tools can lead to bugs, and it is better to use client-go as in the existing way. (For example: we may need to request users to use specific versions of kubectl.) And, client-go has the advantage of type as well.

So, we reached an agreement to use client-go. And it is agreed to implement this import/export feature on web UI as well. 👍

resources/resources.go
```diff
@@ -0,0 +1,194 @@
...
+	// if err := s.schedulerService.RestartScheduler(resources.SchedulerConfig); err != nil {
+	// 	klog.Warningf("failed to start scheduler with imported configuration: %v", err)
+	// 	return nil, xerrors.Errorf("restart scheduler with imported configuration: %w", err)
+	// }
+
+	for _, sc := range *resources.StorageClassList {
+		sc.ObjectMetaApplyConfiguration.UID = nil
+		_, err := s.storageClassService.Apply(ctx, &sc)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to apply StorageClass: %+v", err)
```
sanposhiho, line 146: - Errors should be wrapped with `%w`.
- In this project, error messages contain only "what did it do here". So you don't need `failed to` (This is not a rule in general, but at least in this project it is followed.)

```suggestion
			return nil, xerrors.Errorf("apply StorageClass: %w", err)
```

resources/resources.go
```diff
@@ -0,0 +1,194 @@
...
+	return resources, nil
+}
+
+// Import all resources from posted data.
+// (1) Restart scheduler based on the data.
+// (2) Apply each resource to the scheduler.
+//     * If UID is not nil, an error will occur. (try to find existing resource by UID)
+// (3) Get all resources. (Separated the get function to unify the struct format.)
+func (s *Service) ImportAll(ctx context.Context, resources *ResourcesApplyConfiguration) (*Resources, error) {
+
```
sanposhiho, line 135: I think we can apply all resources in parallel. It makes a better performance.

resources/resources.go
sanposhiho: Hmm, I don't have an idea other than this... 🤔

server/di/interface.go
```diff
@@ -3,6 +3,8 @@ package di
 import (
 	"context"
 
+	"github.com/kubernetes-sigs/kube-scheduler-simulator/resources"
```
sanposhiho, line 6: You should sort import. 
You can do it with `make format`.

resources/resources.go
```diff
@@ -0,0 +1,194 @@
...
+		PodList:          &pods.Items,
+		NodeList:         &nodes.Items,
+		PvList:           &pvs.Items,
+		PvcList:          &pvcs.Items,
+		StorageClassList: &scs.Items,
+		SchedulerConfig:  ss,
+	}, nil
+}
+
+func (s *Service) ExportAll(ctx context.Context) (*Resources, error) {
```
sanposhiho, line 121: nits: Rename this function to `Export`.

resources/resources.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return resources, nil
+}
+
+// Import all resources from posted data.
+// (1) Restart scheduler based on the data.
+// (2) Apply each resource to the scheduler.
+//     * If UID is not nil, an error will occur. (try to find existing resource by UID)
+// (3) Get all resources. (Separated the get function to unify the struct format.)
+func (s *Service) ImportAll(ctx context.Context, resources *ResourcesApplyConfiguration) (*Resources, error) {
```
sanposhiho, line 134: nits: Rename this function to `Import`

resources/resources.go
```diff
@@ -0,0 +1,194 @@
...
+		pvService:           pvs,
+		pvcService:          pvcs,
+		storageClassService: storageClasss,
+		schedulerService:    schedulers,
+	}
+}
+
+// Get all resources from each service.
+func (s *Service) get(ctx context.Context) (*Resources, error) {
+	pods, err := s.podService.List(ctx)
```
sanposhiho, line 89: We can get all resources in parallel for a better performance.

resources/resources.go
```diff
@@ -0,0 +1,194 @@
...
+	nodeService         NodeService
+	pvService           PersistentVolumeService
+	pvcService          PersistentVolumeClaimService
+	storageClassService StorageClassService
+	schedulerService    SchedulerService
+}
+
+// Resources  will used to compile all the resources for export
+type Resources struct {
+	PodList          *[]corev1.Pod                             `json:"podList"`
```
sanposhiho, line 27: Please rename all fields from `HogeList` to `Hoges`. (For example: `PodList` → `Pods`)

sanposhiho: And please change type from `*[]corev1.Pod` to `[]corev1.Pod`. (other fields as well)
I don't see any reasons to use a pointer of slices.

server/handler/resources.go
```diff
@@ -0,0 +1,69 @@
...
+
+type ResourcesHandler struct {
+	service di.ResourcesService
+}
+
+func NewResourcesHandler(s di.ResourcesService) *ResourcesHandler {
+	return &ResourcesHandler{service: s}
+}
+
+func (h *ResourcesHandler) ExportResourcesAll(c echo.Context) error {
```
sanposhiho, line 21: nits: Please rename this function to `Export`

server/handler/resources.go
```diff
@@ -0,0 +1,69 @@
...
+
+	rs, err := h.service.ExportAll(ctx)
+	if err != nil {
+		klog.Errorf("failed to export all resources: %+v", err)
+		return echo.NewHTTPError(http.StatusInternalServerError)
+	}
+	return c.JSON(http.StatusOK, rs)
+}
+
+func (h *ResourcesHandler) ImportResourcesAll(c echo.Context) error {
```
sanposhiho, line 32: nits: Please rename this function to `Import`

server/handler/resources.go
```diff
@@ -0,0 +1,69 @@
...
+	ctx := c.Request().Context()
+
+	// backup before overwrite exist resources.
+	bkp, err := h.service.ExportAll(ctx)
```
sanposhiho, line 36: I think it is better to assume that this import feature can only be used when a user has not created any resources. So, it's not needed to think about existing resources, not needed to take backup.

server/handler/resources.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		if err2 := json.Unmarshal(bkpr, &bkpResources); err2 != nil {
+			klog.Errorf("failed to convert to ResourcesApplyConfiguration of recovery resources: %+v", err2)
+			return echo.NewHTTPError(http.StatusInternalServerError)
+		}
+		_, err3 := h.service.ImportAll(ctx, bkpResources)
+		if err3 != nil {
+			klog.Errorf("failed to recover of backup resources: %+v", err3)
+			return echo.NewHTTPError(http.StatusInternalServerError)
+		}
```
sanposhiho, line 66: Return 500 (HTTPError) when an import fails.

resources/resources.go
```diff
@@ -0,0 +1,194 @@
+package resources
```
sanposhiho, line 1: We cannot understand what this package does from this package name. (`resouces` is too generic a name.)
I think `package export` is better...? I'm not sure, because this service also does import. 😅

server/handler/resources.go
```diff
@@ -0,0 +1,69 @@
...
+	"encoding/json"
+	"net/http"
+
+	"github.com/kubernetes-sigs/kube-scheduler-simulator/resources"
+	"github.com/kubernetes-sigs/kube-scheduler-simulator/server/di"
+	"github.com/labstack/echo/v4"
+	"k8s.io/klog/v2"
+)
+
+type ResourcesHandler struct {
```
sanposhiho, line 13: `ResourcesHandler` is too generic a name as well. It should be renamed like service.

server/handler/resources.go
196Ikuchil: Import is processed in order for each service.
So what should I do if it fails in the middle of the Apply sequence? There are some resources that have already been imported and made. (If there is a way to check if all the imported data is correct before starting Apply...)

resources/resources.go
196Ikuchil: `make format` failed on here.
```
export/export.go:134:1: calculated cyclomatic complexity for function Import is 14, max is 10 (cyclop)
```
What should I do?

196Ikuchil:
TODO List
- [ ] parallel processing on import/export  
- [ ] lint failed on [here](https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/64/commits/2c2972e9e6bec0a54872eceb15994c883dd2ff27#r773230502)
- [ ] test code
- [ ]  what should I do if it fails in the middle of the Apply sequence? [here](https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/64#discussion_r772542003)

export/export.go
```diff
@@ -86,36 +89,78 @@ func NewResourcesService(client clientset.Interface, pods PodService, nodes Node
...
-	}
-	scs, err := s.storageClassService.List(ctx)
-	if err != nil {
-		return nil, xerrors.Errorf("to call list storageClasses")
+	getCtx := context.Background()
+	g, _ := errgroup.WithContext(getCtx)
+	sem := semaphore.NewWeighted(int64(runtime.GOMAXPROCS(0)))
+	resources := Resources{}
+
+	sem.Acquire(getCtx, 1)
```
196Ikuchil, line 97: get lint error with
```
export/export.go:97:13: Error return value of `sem.Acquire` is not checked (errcheck)
	sem.Acquire(getCtx, 1)
```
Should I check in if-statement?

sanposhiho: Yes.

export/export.go
```diff
@@ -0,0 +1,275 @@
...
+		pvService:           pvs,
+		pvcService:          pvcs,
+		storageClassService: storageClasss,
+		schedulerService:    schedulers,
+	}
+}
+
+// Get all resources from each service.
+func (s *Service) get(ctx context.Context) (*Resources, error) {
+	getCtx := context.Background()
```
sanposhiho, line 92: Why do you create new context? (= why don't you use passed `ctx`?)

export/export.go
```diff
@@ -0,0 +1,275 @@
...
+		resources.Pvcs = pvcs.Items
+		return nil
+	})
+
+	sem.Acquire(getCtx, 1)
+	g.Go(func() error {
+		defer sem.Release(1)
+		scs, err := s.storageClassService.List(ctx)
+		if err != nil {
+			return xerrors.Errorf("to call list storageClasses")
```
sanposhiho, line 146: typo?

```suggestion
			return xerrors.Errorf("call list storageClasses")
```

export/export.go
```diff
@@ -0,0 +1,275 @@
...
+	nodeService         NodeService
+	pvService           PersistentVolumeService
+	pvcService          PersistentVolumeClaimService
+	storageClassService StorageClassService
+	schedulerService    SchedulerService
+}
+
+// Resources  will used to compile all the resources for export.
+type Resources struct {
+	Pods            []corev1.Pod                              `json:"podList"`
```
sanposhiho, line 30: nits: `List` is not needed. (other resources as well)
```suggestion
	Pods            []corev1.Pod                              `json:"pods"`
```

export/export.go
```diff
@@ -0,0 +1,275 @@
...
+// (2) Apply each resource to the scheduler.
+//     * If UID is not nil, an error will occur. (try to find existing resource by UID)
+// (3) Get all resources. (Separated the get function to unify the struct format.)
+func (s *Service) Import(ctx context.Context, resources *ResourcesApplyConfiguration) (*Resources, error) {
+	// TODO: Issue: #12 PR: #13
+	// if err := s.schedulerService.RestartScheduler(resources.SchedulerConfig); err != nil {
+	// 	klog.Warningf("failed to start scheduler with imported configuration: %v", err)
+	// 	return nil, xerrors.Errorf("restart scheduler with imported configuration: %w", err)
+	// }
+	getCtx := context.Background()
```
sanposhiho, line 185: same:
> Why do you create new context? (= why don't you use passed ctx?)

server/handler/export.go
```diff
@@ -0,0 +1,47 @@
...
+		klog.Errorf("failed to bind import resources all request: %+v", err)
+		return echo.NewHTTPError(http.StatusInternalServerError)
+	}
+
+	rs, err := h.service.Import(ctx, reqResources)
+	if err != nil {
+		klog.Errorf("failed to import all resources: %+v", err)
+		return echo.NewHTTPError(http.StatusInternalServerError)
+	}
+	return c.JSON(http.StatusOK, rs)
```
sanposhiho, line 46: I think we don't need to return resources after import.

server/handler/resources.go
sanposhiho: > So what should I do if it fails in the middle of the Apply sequence?

I think you don't need to do anything. This will simplify the behavior of the API.
If users want to delete resources that have already been imported, they can use reset resources with reset button. #26

export/export.go
196Ikuchil: I thought that we need to separate the Context. I just didn't understand it...
Fix it now.

export/export.go
196Ikuchil: oops...I'd forgotten...

export/export.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	for i := range resources.Pvs {
+		pv := resources.Pvs[i]
+		if err := sem.Acquire(ctx, 1); err != nil {
+			return xerrors.Errorf("acquire semaphore: %w", err)
+		}
+		g.Go(func() error {
+			defer sem.Release(1)
+			pv.ObjectMetaApplyConfiguration.UID = nil
+			if *pv.Status.Phase == "Bound" {
```
196Ikuchil, line 245: `pv.Status.Phase` is set option to `omitempty`. Therefore, if it is nil, it happens nil pointer error.
In general, no need to be afraid of that. But it may happen if someone rewrites it manually.
How do we check the parameter `Phase` of the struct is exists?

sanposhiho: How about this:
```go
if pv.Status.Phase != nil {
}
```

export/export.go
```diff
@@ -0,0 +1,296 @@
...
+
+// Import all resources from posted data.
+// (1) Restart scheduler based on the data.
+// (2) Apply each resource to the scheduler.
+//     * If UID is not nil, an error will occur. (try to find existing resource by UID)
+// (3) Get all resources. (Separated the get function to unify the struct format.)
+func (s *Service) Import(ctx context.Context, resources *ResourcesApplyConfiguration) error {
+	// TODO: Issue: #12 PR: #13
+	// if err := s.schedulerService.RestartScheduler(resources.SchedulerConfig); err != nil {
+	// 	klog.Warningf("failed to start scheduler with imported configuration: %v", err)
```
sanposhiho, line 200: This warning log is not needed. We usually log errors in handler layers.

server/handler/export.go
```diff
@@ -0,0 +1,47 @@
...
+		klog.Errorf("failed to export all resources: %+v", err)
+		return echo.NewHTTPError(http.StatusInternalServerError)
+	}
+	return c.JSON(http.StatusOK, rs)
+}
+
+func (h *ExportHandler) Import(c echo.Context) error {
+	ctx := c.Request().Context()
+
+	reqResources := new(export.ResourcesApplyConfiguration)
```
sanposhiho, line 35: Please define the same struct as `export.ResourcesApplyConfiguration` in this file, use it to bind requests and convert it into `export.ResourcesApplyConfiguration` to pass struct to service.

Generally, we should not make a handler layer dependent on a service layer, because changing any service layer will cause breaking changes on handler layer. 
In this case, if we change `export.ResourcesApplyConfiguration` in the future, `/import` API will be changed.

export/export.go
196Ikuchil: oops, why couldn't I...thank you

server/handler/export.go
196Ikuchil: So,  in calling of function will be like this?
```
	err := h.service.Import(ctx, &export.ResourcesApplyConfiguration{
		Pods:            reqResources.Pods,
		Nodes:           reqResources.Nodes,
		Pvs:             reqResources.Pvs,
		Pvcs:            reqResources.Pvcs,
		StorageClasses:  reqResources.StorageClasses,
		SchedulerConfig: reqResources.SchedulerConfig,
	})
```

sanposhiho: Yes. (Or, create new function to convert new struct to `export.ResourcesApplyConfiguration`)

196Ikuchil:
/rename Feat: import & export each resource API

196Ikuchil:
/retitle Feat: import & export each resource API

export/export.go
```diff
@@ -195,10 +195,9 @@ func (s *Service) Export(ctx context.Context) (*Resources, error) {
 //     * If UID is not nil, an error will occur. (try to find existing resource by UID)
 // (3) Get all resources. (Separated the get function to unify the struct format.)
 func (s *Service) Import(ctx context.Context, resources *ResourcesApplyConfiguration) error {
-	// TODO: Issue: #12 PR: #13
-	// if err := s.schedulerService.RestartScheduler(resources.SchedulerConfig); err != nil {
-	// 	return nil, xerrors.Errorf("restart scheduler with imported configuration: %w", err)
-	// }
+	if err := s.schedulerService.RestartScheduler(resources.SchedulerConfig); err != nil {
```
196Ikuchil, line 198: I checked this is successful when`pluginConfig` does not exist. 
This issue may be related to #12 and #13.
Therefore, we wait for the merge of the PR.

sanposhiho: Okay. I think we can merge this PR without #13.
I'll review again later.

docs/api.md
```diff
@@ -490,3 +490,62 @@ empty
...
+
+`GET /api/v1/export`
+
+
+### Response
+
+`Content-Type: application/json; charset=UTF-8`
+
+```
+type ResourcesApplyConfiguration struct {
```
sanposhiho, line 508: This should be written in the json format, not the Go format.

docs/api.md
```diff
@@ -490,3 +490,62 @@ empty
...
+
+`POST /api/v1/import`
+
+### Path Parameters
+
+`Content-Type: application/json; charset=UTF-8`
+
+```
+type ResourcesApplyConfiguration struct {
+	Pods            []v1.PodApplyConfiguration                     `json:"pods"`
```
sanposhiho, line 537: same

docs/api.md
```diff
@@ -490,3 +490,62 @@ empty
...
+
+## Import
+
+Import scheduler resources from `json`.
+
+### HTTP Request
+
+`POST /api/v1/import`
+
+### Path Parameters
```
sanposhiho, line 531: It's not path parameters. should be `Request Body`

docs/api.md
196Ikuchil: In that case, the other descriptions also should be fixed to JSON, shouldn't they?
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/4abe7f1946fa69a7abb2243ca08750b4b7eeff76/docs/api.md?plain=1#L28

196Ikuchil: @sanposhiho 
There are a lot of parameters in the upload/download JSON data. 
And this document will be hard to read.

How about saving a sample of the upload/download data in a separate file? (and paste the link in here.)

sanposhiho: > In that case, the other descriptions also should be fixed to JSON, shouldn't they?

It's a fair point. So, okay, for now, let's write it in Go instead of JSON for now. We can improve this doc later if we want.

However, as it is, we don't know which package `v1` or `confstoragev1` represents. So, instead of writing Go struct directly here, I think it would be better to post a link to `ResourcesApplyConfiguration`. (server/handler/export.go:L20)


> How about saving a sample of the upload/download data in a separate file?

Yes. It will be another help for users.

docs/api-samples/v1/export.md
```diff
@@ -0,0 +1,62 @@
...
...Bind":{}},"pluginConfig":[{"name":"DefaultPreemption","args":{"kind":"DefaultPreemptionArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","minCandidateNodesPercentage":10,"minCandidateNodesAbsolute":100}},{"name":"InterPodAffinity","args":{"kind":"InterPodAffinityArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","hardPodAffinityWeight":1}},{"name":"NodeAffinity","args":{"kind":"NodeAffinityArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2"}},{"name":"NodeResourcesBalancedAllocation","args":{"kind":"NodeResourcesBalancedAllocationArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","resources":[{"name":"cpu","weight":1},{"name":"memory","weight":1}]}},{"name":"NodeResourcesFit","args":{"kind":"NodeResourcesFitArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","scoringStrategy":{"type":"LeastAllocated","resources":[{"name":"cpu","weight":1},{"name":"memory","weight":1}]}}},{"name":"PodTopologySpread","args":{"kind":"PodTopologySpreadArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","defaultingType":"System"}},{"name":"VolumeBinding","args":{"kind":"VolumeBindingArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","bindTimeoutSeconds":600}}]}]}}
```
196Ikuchil, line 29: We can also adjust many many indents more sightly. 
But it will not be suited for resending.

196Ikuchil, line 29: @sanposhiho
We can also adjust many many indents more sightly.
But it will not be suited for resending.

sanposhiho (commented):
Please add docs/api-samples/README.md to explain briefly what is these docs.

like:

```
## API samples

These docs are examples of requests and responses of APIs.
```

docs/api-samples/v1/export.md
```diff
@@ -0,0 +1,62 @@
+# Export immediately after startup (empty)
```
sanposhiho, line 1:
```suggestion
# /v1/export samples

## case1: export from a simulator in its initial state
```

docs/api-samples/v1/export.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+# Export resources containing PersistentVolume and PersistentVolumeClaim
```
sanposhiho, line 32:
```suggestion
# case2: export with some PVs and PVCs
```

docs/api-samples/v1/export.md
sanposhiho: I think it's better to format with indents as you said 👍

docs/api-samples/v1/import.md
```diff
@@ -0,0 +1,32 @@
+# Import resources containing PersistentVolume and PersistentVolumeClaim
```
sanposhiho, line 1:
```suggestion
# /v1/import samples

## case1: import resources with some PVs and PVCs
```

docs/api.md
```diff
@@ -490,3 +490,48 @@ empty
...
+`GET /api/v1/export`
+
+
+### Response
+
+`Content-Type: application/json; charset=UTF-8`
+
+`ResourcesApplyConfiguration`(server/handler/export.go:L20)
+
+[Sample data](./api-samples/v1/export.md)
```
sanposhiho, line 509:
```suggestion
You can find sample requests/response [here](./api-samples/v1/export.md)
```

docs/api.md
```diff
@@ -490,3 +490,48 @@ empty
...
+### HTTP Request
+
+`GET /api/v1/export`
+
+
+### Response
+
+`Content-Type: application/json; charset=UTF-8`
+
+`ResourcesApplyConfiguration`(server/handler/export.go:L20)
```
sanposhiho, line 507: nits: let's make it link

```suggestion
[ResourcesApplyConfiguration](server/handler/export.go:L20)
```

docs/api.md
```diff
@@ -490,3 +490,48 @@ empty
...
+
+`POST /api/v1/import`
+
+### Request Body
+
+`Content-Type: application/json; charset=UTF-8`
+
+`ResourcesApplyConfiguration`(server/handler/export.go:L20)
+
+[Sample data](./api-samples/v1/import.md)
```
sanposhiho, line 530: Please add the same changes here.

sanposhiho:
@196Ikuchil 

We support PriorityClasses in #27. So, could you please rebase this branch on latest master and add codes to export/import PriorityClasses? 🙇

196Ikuchil:
Hi @sanposhiho 
> We support PriorityClasses in #27. So, could you please rebase this branch on latest master and add codes to export/import PriorityClasses? 🙇

I corresponded to this. Please check it✋

server/handler/export.go
```diff
@@ -0,0 +1,74 @@
...
+
+	err := h.service.Import(ctx, convertToResourcesApplyConfiguration(reqResources))
+	if err != nil {
+		klog.Errorf("failed to import all resources: %+v", err)
+		return echo.NewHTTPError(http.StatusInternalServerError)
+	}
+	return c.NoContent(http.StatusOK)
+}
+
+// convert from ResourcesApplyConfiguration to export.ResourcesApplyConfiguration.
```
sanposhiho, line 63:
```suggestion
// convertToResourcesApplyConfiguration converts from *ResourcesApplyConfiguration to *export.ResourcesApplyConfiguration.
```

server/handler/export.go
```diff
@@ -0,0 +1,74 @@
...
+	return c.JSON(http.StatusOK, rs)
+}
+
+func (h *ExportHandler) Import(c echo.Context) error {
+	ctx := c.Request().Context()
+
+	reqResources := new(ResourcesApplyConfiguration)
+	if err := c.Bind(reqResources); err != nil {
+		klog.Errorf("failed to bind import resources all request: %+v", err)
+		return echo.NewHTTPError(http.StatusInternalServerError)
```
sanposhiho, line 52: Change the status to bad request.

export/export.go
```diff
@@ -0,0 +1,338 @@
...
+			defer sem.Release(1)
+			pv.ObjectMetaApplyConfiguration.UID = nil
+			if pv.Status != nil && pv.Status.Phase != nil {
+				if *pv.Status.Phase == "Bound" {
+					// PersistentVolumeClaims's UID has been changed to a new value.
+					pvc, err := s.pvcService.Get(ctx, *pv.Spec.ClaimRef.Name)
+					if err == nil {
+						pv.Spec.ClaimRef.UID = &pvc.UID
+					} else {
+						pv.Spec.ClaimRef.UID = nil
```
sanposhiho, line 291: It's helpful to log errors as warn, maybe?

export/export.go
```diff
@@ -0,0 +1,338 @@
...
+		pc := resources.PriorityClasses[i]
+		if err := sem.Acquire(ctx, 1); err != nil {
+			return xerrors.Errorf("acquire semaphore: %w", err)
+		}
+		g.Go(func() error {
+			defer sem.Release(1)
+			pc.ObjectMetaApplyConfiguration.UID = nil
+			_, err := s.priorityclassService.Apply(ctx, &pc)
+			if err != nil {
+				return xerrors.Errorf("apply Node: %w", err)
```
sanposhiho, line 239:
```suggestion
				return xerrors.Errorf("apply PriorityClass: %w", err)
```

docs/api-samples/v1/export.md
```diff
@@ -0,0 +1,65 @@
...
...Bind":{}},"pluginConfig":[{"name":"DefaultPreemption","args":{"kind":"DefaultPreemptionArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","minCandidateNodesPercentage":10,"minCandidateNodesAbsolute":100}},{"name":"InterPodAffinity","args":{"kind":"InterPodAffinityArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","hardPodAffinityWeight":1}},{"name":"NodeAffinity","args":{"kind":"NodeAffinityArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2"}},{"name":"NodeResourcesBalancedAllocation","args":{"kind":"NodeResourcesBalancedAllocationArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","resources":[{"name":"cpu","weight":1},{"name":"memory","weight":1}]}},{"name":"NodeResourcesFit","args":{"kind":"NodeResourcesFitArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","scoringStrategy":{"type":"LeastAllocated","resources":[{"name":"cpu","weight":1},{"name":"memory","weight":1}]}}},{"name":"PodTopologySpread","args":{"kind":"PodTopologySpreadArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","defaultingType":"System"}},{"name":"VolumeBinding","args":{"kind":"VolumeBindingArgs","apiVersion":"kubescheduler.config.k8s.io/v1beta2","bindTimeoutSeconds":600}}]}]}}
```
sanposhiho, line 32: You supported PriorityClasses, so this example should be updated to include PriorityClasses.

sanposhiho: same for `/import` example.

export/export.go
sanposhiho: ^ I think you forgot about this.

sanposhiho (commented):
Thanks for supporting PriorityClasses.
Left a few comments.

docs/api-samples/v1/READMe.md
```diff
@@ -0,0 +1,3 @@
+## API samples
```
sanposhiho, line 1: File name should be `README.md`

docs/api-samples/v1/READMe.md
```diff
@@ -0,0 +1,3 @@
...
+
+These docs are examples of requests and responses of APIs.
```
sanposhiho, line 3:
```suggestion
The sample requests/responses of APIs.
```

docs/api.md
```diff
@@ -487,6 +487,50 @@ delete a priority class.
...
 
 | code  | description |
-| ----- | -------- | 
+| ----- | -------- |
+| 200   | |
+| 500 | something went wrong (see logs of the simulator server) |
+
+## Export
+
+Export current scheduler resources in `json`.
```
sanposhiho, line 496: nits:

```suggestion
Get all resources and current scheduler configuration. 
```

docs/api.md
```diff
@@ -487,6 +487,50 @@ delete a priority class.
...
+
+### HTTP Request
+
+`GET /api/v1/export`
+
+
+### Response
+
+`Content-Type: application/json; charset=UTF-8`
```
sanposhiho, line 505: Delete this line. (We don't write content-type in other APIs)

docs/api.md
```diff
@@ -487,6 +487,50 @@ delete a priority class.
...
+You can find sample requests/responses [here](./api-samples/v1/export.md)
+
+| code  | description |
+| ----- | -------- |
+| 200   | |
+| 500 | something went wrong (see logs of the simulator server) |
+
+## Import
+
+Import scheduler resources from `json`.
```
sanposhiho, line 518: nits:

```suggestion
Apply resources and scheduler configuration.
```

docs/api.md
```diff
@@ -487,6 +487,50 @@ delete a priority class.
...
+
+### HTTP Request
+
+`POST /api/v1/import`
+
+### Request Body
+
+`Content-Type: application/json; charset=UTF-8`
```
sanposhiho, line 526: Delete this line as well.

export/export.go
```diff
@@ -0,0 +1,340 @@
...
+	podService           PodService
+	nodeService          NodeService
+	pvService            PersistentVolumeService
+	pvcService           PersistentVolumeClaimService
+	storageClassService  StorageClassService
+	priorityclassService PriorityClassService
+	schedulerService     SchedulerService
+}
+
+// Resources  will used to compile all the resources for export.
```
sanposhiho, line 40:
```suggestion
// Resources denotes all resources and scheduler configuration for export.
```

export/export.go
```diff
@@ -0,0 +1,340 @@
...
+
+type ResourcesApplyConfiguration struct {
+	Pods            []v1.PodApplyConfiguration                        `json:"pods"`
+	Nodes           []v1.NodeApplyConfiguration                       `json:"nodes"`
+	Pvs             []v1.PersistentVolumeApplyConfiguration           `json:"pvs"`
+	Pvcs            []v1.PersistentVolumeClaimApplyConfiguration      `json:"pvcs"`
+	StorageClasses  []confstoragev1.StorageClassApplyConfiguration    `json:"storageClasses"`
+	PriorityClasses []schedulingcfgv1.PriorityClassApplyConfiguration `json:"priorityClasses"`
+	SchedulerConfig *v1beta2config.KubeSchedulerConfiguration         `json:"schedulerConfig"`
+}
```
sanposhiho, line 59: Let's rename like this to be easier to understand how they use:
- Resources → ResourcesForExport
- ResourcesApplyConfiguration → ResourcesForImport

export/export.go
```diff
@@ -0,0 +1,340 @@
...
+	List(ctx context.Context) (*schedulingv1.PriorityClassList, error)
+	Apply(ctx context.Context, priorityClass *schedulingcfgv1.PriorityClassApplyConfiguration) (*schedulingv1.PriorityClass, error)
+}
+
+type SchedulerService interface {
+	GetSchedulerConfig() *v1beta2config.KubeSchedulerConfiguration
+	RestartScheduler(cfg *v1beta2config.KubeSchedulerConfiguration) error
+}
+
+func NewResourcesService(client clientset.Interface, pods PodService, nodes NodeService, pvs PersistentVolumeService, pvcs PersistentVolumeClaimService, storageClasss StorageClassService, priorityClasss PriorityClassService, schedulers SchedulerService) *Service {
```
sanposhiho, line 97: Rename this to `NewExportService` or `NewExportImportService`

server/di/di.go
```diff
@@ -24,6 +25,7 @@ type Container struct {
 	pvcService           PersistentVolumeClaimService
 	storageClassService  StorageClassService
 	schedulerService     SchedulerService
+	resourcesService     ResourcesService
```
sanposhiho, line 28: Rename this as well. `exportService` or `exportImportService`?

server/di/di.go
```diff
@@ -79,3 +82,7 @@ func (c *Container) SchedulerService() SchedulerService {
 func (c *Container) PriorityClassService() PriorityClassService {
 	return c.priorityClassService
 }
+
```
sanposhiho, line 85: add doc for `ResourcesService` like other services.

196Ikuchil:
@sanposhiho  Thank you for your comments! I fixed it.

sanposhiho (commented):
Thanks! only small comments!

Also I think you miss this comment:
https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/64/files#r785405682

sanposhiho (commented):
All changes looks ok to me! 👍 

Could you fix issues from linters?
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kube-scheduler-simulator/64/pull-kube-scheduler-simulator-backend-lint/1484824816627224576/build-log.txt

Also, you can run lint by `make lint` command.

196Ikuchil:
@sandipanpanda 
Thank you!

> Could you fix issues from linters?

I don't want to divide the existing function's code into shorter ones. 
May I use this comment?
```
//nolint:funlen 
```

`make lint`&`make format` have already run👍

sanposhiho:
> May I use this comment?

Hmm, I would say no. 
Could you tell me why do you think it's better not to split into small funcs?

196Ikuchil:
@sanposhiho 
If divide the function's code, I should do it for each method call of each service, I think. 
But those divided functions may be called only once in current logic. It also distributes where the code is written. So, this division is only for following the lint rule and I don't see the reason to do that.

But I'm not the one who makes the rule of lint and I don't know the objective of the lint rule. So I think it is also good to follow the rule. 
Let me know the objective of the lint rule, please?

sanposhiho:
To express what you are saying in an extreme way: "if no reason, it is better to put all logics into one function even if it will be big." Hmm bro I'd like to say it's a big No.

Note that splitting a big function into small functions is not only for making codes DRY, but also for other many benefits like readability, ease of testing, ease of changing logic... etc. And if not only for now, but also for the future.

So, as you said, if I saw only the current `Import` and `get`, I would feel it's easy enough to read and understand and not need to split. However, our project is expected to grow bigger and bigger, and, this `Import` and `get` functions also should glow bigger and bigger and may have more complicated logics. We need to see the future and need to split such big functions if possible.

196Ikuchil:
Understood it, thank you!
I'll try it, later. please wait a few days..👌

export/export.go
```diff
@@ -108,185 +108,239 @@ func NewExportService(client clientset.Interface, pods PodService, nodes NodeSer
...
+// (3) Get all resources. (Separated the get function to unify the struct format.)
+func (s *Service) Import(ctx context.Context, resources *ResourcesForImport) error {
+	if err := s.schedulerService.RestartScheduler(resources.SchedulerConfig); err != nil {
+		return xerrors.Errorf("restart scheduler with imported configuration: %w", err)
+	}
+	errgrp := NewErrorGroupWithSemaphore(ctx)
+	errgrp.applyPcs(ctx, &s.priorityclassService, resources)
+	errgrp.applyStorageClasses(ctx, &s.storageClassService, resources)
+	errgrp.applyPvcs(ctx, &s.pvcService, resources)
+	errgrp.applyPvs(ctx, &s.pvService, &s.pvcService, resources)
```
196Ikuchil, line 165: I noticed that there is a possibility to fail to establish a relationship with PersistentVolume and PersistentVolumeClaim when PersistentVolumeClaim hasn't been applied yet.

So, applyPvs() should wait for a finish of applyPvcs(). 
Would you give me some idea to implement another asynchronous processing on it.

196Ikuchil:
Hi! @sanposhiho 
> I'll try it, later. please wait a few days..👌

I finished. But I noticed a small bug...
Please check my above comment.

export/export.go
```diff
@@ -172,6 +171,37 @@ func (s *Service) Import(ctx context.Context, resources *ResourcesForImport) err
...
+	if err := eg.sem.Acquire(ctx, 1); err != nil {
+		klog.Fatalf("failed to acquire semaphore: %v", err)
+		return
+	}
+	eg.g.Go(func() error {
+		defer eg.sem.Release(1)
+		errgrp.applyPvcs(ctx, &s.pvcService, resources)
+		return nil
+	})
+	if err := eg.g.Wait(); err != nil {
```
196Ikuchil, line 187: Is this look good?

export/export.go
```diff
@@ -0,0 +1,438 @@
...
+		klog.Fatalf("failed to apply pvcs: %v", err.Error())
+	}
+}
+
+func (eg *ErrorGroupWithSemaphore) listPods(ctx context.Context, s *PodService, r *ResourcesForExport) {
+	if err := eg.sem.Acquire(ctx, 1); err != nil {
+		klog.Fatalf("failed to acquire semaphore: %v", err)
```
sanposhiho, line 207: return error instead of `klog.Fatalf`. `klog.Fatalf` calls `os.Exit` internally. Other places as well.

`os.Exit` shouldn't be used in most cases. (= `klog.Fatalf` also shouldn't be used)

export/export.go
```diff
@@ -0,0 +1,438 @@
...
+		schedulerService:     schedulers,
+	}
+}
+
+type ErrorGroupWithSemaphore struct {
+	g   *errgroup.Group
+	sem *semaphore.Weighted
+}
+
+func NewErrorGroupWithSemaphore(ctx context.Context) ErrorGroupWithSemaphore {
```
sanposhiho, line 116: This `ErrorGroupWithSemaphore` is a nice idea 👍  
Could you move this to the util package so that we can use it from other packages?

https://github.com/kubernetes-sigs/kube-scheduler-simulator/tree/master/util

sanposhiho: And one nits: rename this from `ErrorGroupWithSemaphore` to `ErrGroupWithSemaphore`

export/export.go
```diff
@@ -0,0 +1,432 @@
...
+type ResourcesForImport struct {
+	Pods            []v1.PodApplyConfiguration                        `json:"pods"`
+	Nodes           []v1.NodeApplyConfiguration                       `json:"nodes"`
+	Pvs             []v1.PersistentVolumeApplyConfiguration           `json:"pvs"`
+	Pvcs            []v1.PersistentVolumeClaimApplyConfiguration      `json:"pvcs"`
+	StorageClasses  []confstoragev1.StorageClassApplyConfiguration    `json:"storageClasses"`
+	PriorityClasses []schedulingcfgv1.PriorityClassApplyConfiguration `json:"priorityClasses"`
+	SchedulerConfig *v1beta2config.KubeSchedulerConfiguration         `json:"schedulerConfig"`
+}
+type ErrGroupWithSemaphore util.ErrGroupWithSemaphore
```
sanposhiho, line 60: I think this type alias is not needed. 
Let's remove this and use `util.ErrGroupWithSemaphore` directly.  
`listHoge` and `applyHoge` currently use `*ErrGroupWithSemaphore` as method receiver; change these functions to receive `*util.ErrGroupWithSemaphore` in args.

export/export.go
```diff
@@ -0,0 +1,432 @@
...
+		pvcService:           pvcs,
+		storageClassService:  storageClasss,
+		priorityclassService: priorityClasss,
+		schedulerService:     schedulers,
+	}
+}
+
+func NewErrGroupWithSemaphore(ctx context.Context) ErrGroupWithSemaphore {
+	return ErrGroupWithSemaphore(util.NewErrGroupWithSemaphore(ctx))
+}
```
sanposhiho, line 113: Remove this as well. Replace all `NewErrGroupWithSemaphore` usages with `util.NewErrGroupWithSemaphore`

export/export.go
```diff
@@ -0,0 +1,432 @@
...
+	if err := errgrp.applyPods(ctx, &s.podService, resources); err != nil {
+		return xerrors.Errorf("call applyPods: %w", err)
+	}
+
+	if err := errgrp.Grp.Wait(); err != nil {
+		return xerrors.Errorf("apply each resources: %w", err)
+	}
+
+	// The apply of PersistentVolume will be called after the apply of PersistentVolumeClaim is finish.
+	// This is because that PersistentVolume search PersistentVolumeClaim and associate itself if they had relationship.
```
sanposhiho, line 188:
```suggestion
	// `applyPvs` should be called after `applyPvcs` finished,
	// because `applyPvs` look up PersistentVolumeClaim for `Spec.ClaimRef.UID` field.
```

export/export.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+	if err := errgrp.applyNodes(ctx, &s.nodeService, resources); err != nil {
+		return xerrors.Errorf("call applyNodes: %w", err)
+	}
+	if err := errgrp.applyPods(ctx, &s.podService, resources); err != nil {
+		return xerrors.Errorf("call applyPods: %w", err)
+	}
+
+	if err := errgrp.Grp.Wait(); err != nil {
+		return xerrors.Errorf("apply each resources: %w", err)
```
sanposhiho, line 184:
```suggestion
		return xerrors.Errorf("apply resources: %w", err)
```

export/export.go
```diff
@@ -0,0 +1,432 @@
...
+
+func (s *Service) Export(ctx context.Context) (*ResourcesForExport, error) {
+	resources, err := s.get(ctx)
+	if err != nil {
+		return nil, xerrors.Errorf("export resources all: %w", err)
+	}
+	return resources, nil
+}
+
+// Import all resources from posted data.
```
sanposhiho, line 156:
```suggestion
// Import imports all resources from posted data.
```

export/export.go
```diff
@@ -0,0 +1,432 @@
...
+		priorityclassService: priorityClasss,
+		schedulerService:     schedulers,
+	}
+}
+
+func NewErrGroupWithSemaphore(ctx context.Context) ErrGroupWithSemaphore {
+	return ErrGroupWithSemaphore(util.NewErrGroupWithSemaphore(ctx))
+}
+
+// Get all resources from each service.
```
sanposhiho, line 115:
```suggestion
// Get gets all resources from each service.
```

export/export.go
```diff
@@ -0,0 +1,432 @@
...
+		return xerrors.Errorf("call applyPvs: %w", err)
+	}
+	if err := errgrp.Grp.Wait(); err != nil {
+		return xerrors.Errorf("apply applyPvs: %w", err)
+	}
+
+	return nil
+}
+
+func (eg *ErrGroupWithSemaphore) listPods(ctx context.Context, s *PodService, r *ResourcesForExport) error {
```
sanposhiho, line 199: All `listHoge` and `applyHoge` receive HogeService as pointer. It isn't needed to be pointer. 
With this change, we can also change `(*s).List(ctx)` in L205 to `s.List(ctx)`.

```suggestion
func (eg *ErrGroupWithSemaphore) listPods(ctx context.Context, s PodService, r *ResourcesForExport) error {
```

sanposhiho (commented):
Thanks

export/export.go
196Ikuchil: > With this change, we can also change (*s).List(ctx) in L205 to s.List(ctx).

Due to the last fix, I will use `Service` as method receiver.

export/export.go
```diff
@@ -0,0 +1,427 @@
...
+	resources, err := s.get(ctx)
+	if err != nil {
+		return nil, xerrors.Errorf("export resources all: %w", err)
+	}
+	return resources, nil
+}
+
+// Import imports all resources from posted data.
+// (1) Restart scheduler based on the data.
+// (2) Apply each resource to the scheduler.
```
sanposhiho, line 153: Not `to scheduler`, but `to api-server`. I think just deleting `to the scheduler` is fine.

```suggestion
// (2) Apply each resource.
```

export/export.go
```diff
@@ -0,0 +1,427 @@
...
+//     * If UID is not nil, an error will occur. (try to find existing resource by UID)
+// (3) Get all resources. (Separated the get function to unify the struct format.)
```
sanposhiho, line 155: Which action is this referring to? an out-of-date comment, maybe?

export/export.go
```diff
@@ -0,0 +1,427 @@
...
+		return xerrors.Errorf("apply resources: %w", err)
+	}
+
+	// `applyPvs` should be called after `applyPvcs` finished,
+	// because `applyPvs` look up PersistentVolumeClaim for `Spec.ClaimRef.UID` field.
+	if err := s.applyPvs(ctx, resources, &errgrp); err != nil {
+		return xerrors.Errorf("call applyPvs: %w", err)
+	}
+	if err := errgrp.Grp.Wait(); err != nil {
+		return xerrors.Errorf("apply applyPvs: %w", err)
```
sanposhiho, line 188: nits: 

```suggestion
		return xerrors.Errorf("apply PVs: %w", err)
```

sanposhiho (commented):
Thanks. Just three nits!

export/export.go
196Ikuchil: Yes, that process has already been removed. 
I will delete this line.

sanposhiho (approved):
Thanks for the long journey...! Approve.

/lgtm
/approve

PR #54, opened 2021-12-13
add reset scheduler configuration

sivchari:
#### What type of PR is this?
/kind feature
/kind api-change

#### What this PR does / why we need it:
Add api to reset scheduler configuration.

#### Which issue(s) this PR fixes:

Related #26

#### Special notes for your reviewer:

/label tide/merge-method-squash

PR #70, opened 2021-12-21
Fix reset scheduler configuration endpoint path

sanposhiho:
#### What type of PR is this?

/kind api-change

We cannot guess what `PUT /api/v1/schedulerconfiguration` will do. 
(I should've pointed it when I reviewed #54 😓)

This PR change the endpoint to `PUT /api/v1/schedulerconfiguration/reset`

Ref #26 

This PR is not needed in case that we re-design #26 to create one API to reset all resources and scheduler config.
So, let me `/hold` this. 

/cc @sivchari 
/hold

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -44,7 +44,7 @@ restart scheduler with default configuration.
 
 ### HTTP Request
 
-`PUT /api/v1/schedulerconfiguration`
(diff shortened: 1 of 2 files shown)


sanposhiho:
We re-designed #26. We will delete this API and create one API to reset all resources and scheduler config.
This PR is no longer needed.

/close

PR #75, opened 2021-12-26
enable admission control with `Priority` admission plugin to support Preemption

### LINKED ISSUE #74: enable admission control with `Priority` admission plugin to support Preemption
SSIA. 

We have to enable `Priority` admission plugin to handle PriorityClass properly. 
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#priority

/assign

sanposhiho:
#### What type of PR is this?

/kind feature

- enable admission control with `Priority` admission plugin to support Preemption
- change terminationGracePeriodSeconds on template to 0. `terminationGracePeriodSeconds` must be 0 to be preempted on simulator. This is because we don't run kubelet on simulator and if `terminationGracePeriodSeconds` != 0, the Pod isn't deleted forever. (I will add this behavior to the doc later.)

Fixes #74

### DIFF
--- a/k8sapiserver/k8sapiserver.go
+++ b/k8sapiserver/k8sapiserver.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/kubernetes/pkg/controlplane"
 	"k8s.io/kubernetes/pkg/kubeapiserver"
 	kubeletclient "k8s.io/kubernetes/pkg/kubelet/client"
+	"k8s.io/kubernetes/plugin/pkg/admission/priority"
 
 	generated "github.com/kubernetes-sigs/kube-scheduler-simulator/k8sapiserver/openapi"
 )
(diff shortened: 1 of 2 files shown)


sanposhiho:
With this PR, we finally support Preemption on simulator 🎉 

1. create two low-priority Pods.
スクリーンショット 2021-12-26 13 59 16

2. create one high-priority Pod.
3. One low-priority Pod is preempted.
スクリーンショット 2021-12-26 13 59 41

4. High-priority Pod is bound to nominated Node.
スクリーンショット 2021-12-26 14 00 14

aibarbetta:
Hi @sanposhiho, I'm running this branch of the simulator on my machine and trying [this](https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/75#issuecomment-1001107992) example. I've noticed I have to apply the `priority-high-pod` a second time for it to bind to the nominated `node1`, is this by design? Just asking because I was expecting the pod to appear in the node automatically

sanposhiho:
@aibarbetta Sorry, I missed your comment.

Maybe you didn't need to apply twice, just reloading the window should be fine so that the web client re-fetch the resources.
The scheduler is scheduling Pods asynchronously and scheduling for priority-high-pod should be run twice. (because priority-high-pod needs to preempt another pod to be bound. It's due to how preemption works.) Web client re-fetches the resources just after applying the resources so scheduling may not be finished before the client fetches the resource.

PR #88, opened 2022-02-01
fix(frontend): react on nuxt telemetry user input during `yarn start`

### LINKED ISSUE #87: Bug: `frontend` is not available, due to missing `nuxt` telemetry user input during `yarn start`
While trying to run this application via `docker-compose` I noticed that the `frontend` is not available via http://localhost:3000.
I did some evaluation and noticed that there is a `nuxt` telemetry user-input required when starting the `frontend` application.

```bash
> podman exec -it simulator-frontend

/app $ netstat -an -t
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State

/app $ yarn start
yarn run v1.22.15
$ nuxt start

ℹ NuxtJS collects completely anonymous data about usage.
  This will help us improve Nuxt developer experience over time.
  Read more on https://git.io/nuxt-telemetry

? Are you interested in participating? (Y/n)
```

After some digging I found the following input on that:
https://github.com/nuxt/telemetry#opting-out

When adding the described env variable and rebuilding the `frontend` image, the application is starting without problems and directly available:

```bash
> podman exec -it simulator-frontend

/app $ netstat -an -t
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       
tcp        0      0 0.0.0.0:3000            0.0.0.0:*               LISTEN
```

I will create a PR for that change

skuethe:
Signed-off-by: skuethe <56306041+skuethe@users.noreply.github.com>

#### What type of PR is this?

/kind bug

This PR adds the required `nuxt` telementry user input via env variable during the container image build. Without this input

Fixes #87 

There are two possibilities to avoid the telemetry user input during startup:
https://github.com/nuxt/telemetry#opting-out

1. modify a setting in a `nuxt.config`
2. set an environment var

I tried to modify the file `web/nuxt.config.js` and then re-enable to copy command in the `web/Dockerfile`, but that does seem to create a lot of other issues.
So this PR is setting the env var in the `web/Dockerfile`.

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -18,6 +18,7 @@ FROM node:16-alpine AS runner
 WORKDIR /app
 
 ENV NODE_ENV production
+ENV NUXT_TELEMETRY_DISABLED 1
 
 RUN addgroup -g 1001 -S nodejs
 RUN adduser -S nuxtjs -u 1001


sanposhiho (commented):
Looks good. Thanks 👍 

/lgtm
/approve

Issue #41, opened 2021-11-06
Feature: Simulate Pods scheduling process in any existing clusters


lianghao208:
Hi, I am new to this project (from scheduling-plugin sig). I learned about the kube-scheduler-simulator. In my option, this simulator only simulate pods scheduling process in a new simulation cluster, which means it can't simulate a real cluster for now(e.g production environment) . Is it possible to simulate pods in an already exist environment?
I am interesting in contributing to this project, maybe I need to learn more details about it.

sanposhiho:
Hi @lianghao208 :)

Currently, when you start the simulator, there are no resources on the simulator, and we have to create resources(pods, nodes, ...) one by one to simulate cluster. 

#28 is one idea to enable user to export current resources **on the simulator** to some file and import resources with that file later. 
But your idea is importing resources from **a real cluster**, right? That makes sense to me.

> I am interesting in contributing to this project, maybe I need to learn more details about it.

Yes, you can work on this feature if you want. I will support you as much as I can.

lianghao208:
> But your idea is importing resources from a real cluster, right? That makes sense to me :)

Yes, my idea is to import a real cluster's etcd data (etcd's endpoint, certs are needed) into the simulator, then the simulator will create the same resources as the imported cluster's.

> Yes, you can work on this feature if you want. I will support you as much as I can.

Thank you, I will try to work on this.

sanposhiho:
> my idea is to import a real cluster's etcd data (etcd's endpoint, certs are needed) into the simulator, then the simulator will create the same resources as the imported cluster's.

I see.

Let me ask you one question. Can't most users get certs of etcd easily because getting etcd certs requires strong permissions in the cluster? Do you expect the administrator to use this export feature? Or is there an easy way for users to get the cert for the real cluster etcd? (I'm not very familiar with etcd certificate in Kubernetes.)

I think there are several ways to achieve this import feature, including your way. And the way I'm thinking is to get resources from api-server of real cluster. 

> I will try to work on this.

Thanks. Feel free to `/assign` :)

flowscolors:
[...]

PR #97, opened 2022-02-08
Prep work for import function from existing clusters

196Ikuchil:
#### What type of PR is this?

/kind api-change

Eventually, we would like to implement #41 . This is prep work.
- Add an option for the simulator to enable external import feature.
- Add a new option to `ignore error`on the argument of import/export functions.

Relates #41

This PR is the idea of me. So, this might be rechanged all.

/hold

### DIFF
--- a/export/export.go
+++ b/export/export.go
@@ -107,28 +107,46 @@ func NewExportService(client clientset.Interface, pods PodService, nodes NodeSer
 	}
 }
 
+type options struct {
+	ignoreErr bool
+}
+
+type ignoreErrOption bool
+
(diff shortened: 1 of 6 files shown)


export/export.go
```diff
@@ -107,28 +107,46 @@ func NewExportService(client clientset.Interface, pods PodService, nodes NodeSer
 	}
 }
 
+type options struct {
```
sanposhiho, line 110: Thanks for taking my suggestion for the functional options pattern. Looks nice. :)

PR #103, opened 2022-02-11
Support to `WithIgnoreErr` option in the `ExportService`

196Ikuchil:
#### What type of PR is this?

/kind feature
/kind api-change

Support to `WithIgnoreErr` option in the export service. 

Eventually, we would like to implement https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/41 .
> When this is passed, `export/import` method shouldn't return errors and log the error instead.
> - The simulator may only be able to get some of the resources due to lack of permissions or other reasons.
> - And some users may want to import just some of the resources from an existing cluster.

Relates #41

### DIFF
--- a/export/export.go
+++ b/export/export.go
@@ -107,28 +107,46 @@ func NewExportService(client clientset.Interface, pods PodService, nodes NodeSer
 	}
 }
 
+type options struct {
+	ignoreErr bool
+}
+
+type ignoreErrOption bool
+
+type Option interface {
+	apply(*options)
+}
+
+func (i ignoreErrOption) apply(opts *options) {
+	opts.ignoreErr = bool(i)
+}
+
+func (s *Service) WithIgnoreErr() Option {
+	return ignoreErrOption(true)
+}
+
 // Get gets all resources from each service.
-func (s *Service) get(ctx context.Context) (*ResourcesForExport, error) {
+func (s *Service) get(ctx context.Context, opts options) (*ResourcesForExport, error) {
 	errgrp := util.NewErrGroupWithSemaphore(ctx)
 	resources := ResourcesForExport{}
 
-	if err := s.listPods(ctx, &resources, &errgrp); err != nil {
+	if err := s.listPods(ctx, &resources, &errgrp, opts); err != nil {
 		return nil, xerrors.Errorf("call listPods: %w", err)
 	}
-	if err := s.listNodes(ctx, &resources, &errgrp); err != nil {
+	if err := s.listNodes(ctx, &resources, &errgrp, opts); err != nil {
 		return nil, xerrors.Errorf("call listNodes: %w", err)
 	}
-	if err := s.listPvs(ctx, &resources, &errgrp); err != nil {
+	if err := s.listPvs(ctx, &resources, &errgrp, opts); err != nil {
 		return nil, xerrors.Errorf("call listPvs: %w", err)
 	}
-	if err := s.listPvcs(ctx, &resources, &errgrp); err != nil {
+	if err := s.listPvcs(ctx, &resources, &errgrp, opts); err != nil {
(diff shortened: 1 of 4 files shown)


export/export.go
```diff
@@ -180,105 +196,146 @@ func (s *Service) Import(ctx context.Context, resources *ResourcesForImport) err
...
 		pods, err := s.podService.List(ctx)
 		if err != nil {
-			return xerrors.Errorf("call list pods: %w", err)
+			if !opts.ignoreErr {
+				return xerrors.Errorf("call list pods: %w", err)
+			}
+			klog.Errorf("failed to call list pods: %v", err)
+			pods = &corev1.PodList{Items: []corev1.Pod{}}
 		}
 		r.Pods = pods.Items
```
196Ikuchil, line 240: I added this line.
```go
pods = &corev1.PodList{Items: []corev1.Pod{}}
```

If `s.podService.List(ctx)` return an error, `pods` is going to `nil`.
So, runtime error has occurred in here.

export/export.go
```diff
@@ -107,28 +107,46 @@ func NewExportService(client clientset.Interface, pods PodService, nodes NodeSer
...
+
+type Option interface {
+	apply(*options)
+}
+
+func (i ignoreErrOption) apply(opts *options) {
+	opts.ignoreErr = bool(i)
+}
+
+func (s *Service) WithIgnoreErr() Option {
```
sanposhiho, line 124: nits: Rename this func to just `IgnoreErr`. And add the doc for this like:

```suggestion
// IgnoreErr is the option to literally ignore errors.
// If it is enabled, the method won't return any errors, but just log errors as error logs.
func (s *Service) IgnoreErr() Option {
```

export/export.go
```diff
@@ -107,28 +107,46 @@ func NewExportService(client clientset.Interface, pods PodService, nodes NodeSer
...
 		return nil, xerrors.Errorf("call listPvcs: %w", err)
 	}
-	if err := s.listStorageClasses(ctx, &resources, &errgrp); err != nil {
+	if err := s.listStorageClasses(ctx, &resources, &errgrp, opts); err != nil {
 		return nil, xerrors.Errorf("call listStorageClasses: %w", err)
 	}
-	if err := s.listPcs(ctx, &resources, &errgrp); err != nil {
-		return nil, xerrors.Errorf("call listPcs: %w", err)
+	if err := s.listPcs(ctx, &resources, &errgrp, opts); err != nil {
+		return nil, xerrors.Errorf("call Pcs: %w", err)
```
sanposhiho, line 149: This change looks mistake, maybe? I don't see any reason for this. (because other formats of error are `call listHoge: %w`)

export/export.go
```diff
@@ -140,38 +158,36 @@ func (s *Service) get(ctx context.Context) (*ResourcesForExport, error) {
...
-		return xerrors.Errorf("restart scheduler with imported configuration: %w", err)
-	}
+// Apply applies all resources from each service.
+func (s *Service) apply(ctx context.Context, resources *ResourcesForImport, opts options) error {
 	errgrp := util.NewErrGroupWithSemaphore(ctx)
 
-	if err := s.applyPcs(ctx, resources, &errgrp); err != nil {
-		return xerrors.Errorf("call applyPcs: %w", err)
+	if err := s.applyPcs(ctx, resources, &errgrp, opts); err != nil {
+		return err
```
sanposhiho, line 178: Let's wrap errors as before.

sanposhiho (commented):
Looks super good. Nice job 👍 
Left a few small comments.

196Ikuchil:
@sanposhiho Thanks! Fixed it.

sanposhiho (approved):
Thanks. Looks good :)

sanposhiho (commented):
/lgtm

PR #111, opened 2022-02-16
Import from existing cluster

### LINKED ISSUE #41: Feature: Simulate Pods scheduling process in any existing clusters

196Ikuchil:
#### What type of PR is this?
/kind feature

The simulator will be able to import resources from an existing cluster indicated by a `kubeconfig` file. 
`kubeconfig` is indicated by `KUBECONFIG` environment variable.
That import logic is enabled when environment variable `EXTERNAL_IMPORT_ENABLED` is `"1"`.

【new feature】
- `kubeconfig` will be imported when the simulator is launched.
- `ignoreRestart` Option is implemented to `ExportService` for the `Import` function.
- Implement `ReplicateExistingClusterService` that has two `ExportService`.
  -  The one is used for exporting the resources from the existing cluster.
  - Another one is used for importing those to the simulator.

Fixes #41

The scheduler service of this simulator is not restarting when the `Import` function is executed. 
This means that if we set `kube scheduler config file` of the existing cluster to the `KUBE_SCHEDULER_CONFIG_PATH` environment variable, we can apply the `kube scheduler config file` to the simulator. (we can't read the `kube scheduler config` from the existing cluster.)

About the `import` flow.
1. Do `yarn start`.
2. Read `KUBE_SCHEDULER_CONFIG_PATH`  and `KUBECONFIG` environment variables and import the file.
3. Create many `Services`.
4. Start `Scheduler`. (this scheduler is used for a long time  after this.)
5. Do `Import` function of this PR and import the existing cluster's resources to the scheduler.
6. Finish.

I tested this function for each resource on local and the import was successful.

- [x] WIP:some test code.

/hold


replicateexistingcluster/replicateexistingcluster.go
```diff
@@ -0,0 +1,185 @@
+package replicateexistingcluster
```
196Ikuchil, line 1: This name is too long?

replicateexistingcluster/replicateexistingcluster.go
```diff
@@ -0,0 +1,185 @@
...
+	expRes, err := s.existingClusterExportService.Export(ctx)
+	if err != nil {
+		return xerrors.Errorf("call Export of existingClusterExportService: %w", err)
+	}
+	impRes, err := cnvtToResourcesForImportFromResourcesForExport(expRes)
+	if err != nil {
+		return xerrors.Errorf("call cnvtToResourcesForImportFromResourcesForExport: %w", err)
+	}
+	// Import to the simulator without restarting it.
+	if err := s.simulatorExportService.Import(ctx, impRes, s.simulatorExportService.IgnoreErr(), s.simulatorExportService.IgnoreRestart()); err != nil {
```
196Ikuchil, line 47: Call the `Import` function of the `ExportService` to the simulator.
And it will not restart the simulator.

replicateexistingcluster/replicateexistingcluster.go
```diff
@@ -0,0 +1,185 @@
...
+		Pvs:             pvs,
+		Pvcs:            pvcs,
+		StorageClasses:  scs,
+		PriorityClasses: pcs,
+		// existingClusterExportService can't export the SchedulerConfig.
+		SchedulerConfig: nil,
+	}, nil
+}
+
+func cnvtPodListToApplyConfigurationList(pods []corev1.Pod) ([]v1.PodApplyConfiguration, error) {
```
196Ikuchil, line 90: Is this good for this converting logic?
Actually, I want to convert a method like this.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/2802ed2a60aa6c168e21c44f42ffcd52a7ac05a2/scheduler/scheduler.go#L169

But I can't find it. Please give me some good idea if you have?

lianghao208: Yes, it's a good idea to use a generic conversion functions to convert internal object to applyConfiguration object.

Maybe you can refer to this extractor code:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/pod-security-admission/admission/admission.go#L111-L134

In this case, it becomes:
```go
func convertToApplyConfiguration(in interface{}) (interface{}, error) {
	_in, err := json.Marshal(in)
         if err != nil {
		return nil, xerrors.Errorf("call Marshal to cnvt object: %w", err)
	}
	switch in.(type) {
	case *corev1.Pod:
		p := &applyv1.PodApplyConfiguration{}
		if err := json.Unmarshal(_in, &p); err != nil {
			return nil, xerrors.Errorf("call Marshal to cnvt Pod: %w", err)
		}
		return p, nil
        case *corev1.Node:
                ...
        case *corev1.PersistentVolume:
                ...
        case *corev1.PersistentVolumeClaim:
                ...
	default:
		return nil, xerrors.Errorf("unknown type")
	}
	return nil
}
```
@sanposhiho WDYT?

sanposhiho (commented):
This time I only reviewed the comments and wording. 
I'll review implementation details this weekend.

replicateexistingcluster/replicateexistingcluster.go
196Ikuchil: Thank you for both advice!
I am going to implement it based on @sanposhiho 's one.
it is also good for testing, I think.

replicateexistingcluster/replicateexistingcluster.go
```diff
@@ -0,0 +1,224 @@
...
+	rto := make([]schedulingcfgv1.PriorityClassApplyConfiguration, len(pcs))
+	for i, p := range pcs {
+		if err := convertToApplyConfiguration(p, &rto[i]); err != nil {
+			return nil, xerrors.Errorf("convert PriorityClasses to apply configuration: %w", err)
+		}
+	}
+	return rto, nil
+}
+
+func convertToApplyConfiguration(in interface{}, out interface{}) error {
```
sanposhiho, line 161:
```suggestion
// convertToApplyConfiguration is convert some object to XXXXApplyConfiguration.
// out should be the pointer of XXXXApplyConfiguration, otherwise you cannot get result of conversion.
//nolint:funlen // For readability.
func convertToApplyConfiguration(in interface{}, out interface{}) error {
```

replicateexistingcluster/replicateexistingcluster.go
```diff
@@ -0,0 +1,224 @@
...
+	rto := make([]v1.PersistentVolumeClaimApplyConfiguration, len(pvcs))
+	for i, p := range pvcs {
+		if err := convertToApplyConfiguration(p, &rto[i]); err != nil {
+			return nil, xerrors.Errorf("convert PersistentVolumeClaim to apply configuration: %w", err)
+		}
+	}
+	return rto, nil
+}
+
+func cnvtStorageClassesListToApplyConfigurationList(scs []storagev1.StorageClass) ([]cfgstoragev1.StorageClassApplyConfiguration, error) {
```
sanposhiho, line 141: Could you replace all `cnvt` with `convert`?

replicateexistingcluster/replicateexistingcluster.go
```diff
@@ -0,0 +1,224 @@
...
+		return xerrors.Errorf("call cnvtToResourcesForImportFromResourcesForExport: %w", err)
+	}
+	// Import to the simulator.
+	if err := s.simulatorExportService.Import(ctx, impRes, s.simulatorExportService.IgnoreErr(), s.simulatorExportService.IgnoreSchedulerConfiguration()); err != nil {
+		return xerrors.Errorf("call Import of the simulater export service: %w", err)
+	}
+	return nil
+}
+
+func cnvtToResourcesForImportFromResourcesForExport(expRes *export.ResourcesForExport) (*export.ResourcesForImport, error) {
```
sanposhiho, line 64: I think these convert logic should be provided in `export` package.
Maybe, you might want to create a new file `util.go` in this package which includes all utility funcs for export service.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/tree/master/export

196Ikuchil: Thanks, good idea!

196Ikuchil:
Thanks for the review✋ Fixed it.

export/utils.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return &ResourcesForImport{
+		Pods:            pods,
+		Nodes:           nodes,
+		Pvs:             pvs,
+		Pvcs:            pvcs,
+		StorageClasses:  scs,
+		PriorityClasses: pcs,
+		// existingClusterExportService can't export the SchedulerConfig.
+		SchedulerConfig: nil,
```
sanposhiho, line 48: Now, this `ConvertToResourcesForImportFromResourcesForExport` is utility func. So, set `expRes.SchedulerConfig` here.
In replicateexistingcluster service, `ResourcesForExport.SchedulerConfig` from existing cluster should be already nil. So, just changing here as the above will work well.

replicateexistingcluster/replicateexistingcluster.go
```diff
@@ -0,0 +1,185 @@
...
+	return &Service{
+		existingClusterExportService: existingClusterExportService,
+		simulatorExportService:       exportService,
+	}
+}
+
+// ImportFromExistingCluster get resources from existing cluster via existingClusterExportService
+// and then apply those to the simulator without restarting it via simulatorExportService.
+func (s *Service) ImportFromExistingCluster() error {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
sanposhiho: Please do this.

sanposhiho (commented):
Thanks. Just two nits!

export/utils.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, xerrors.Errorf("call convertPriorityClassesListToApplyConfigurationList: %w", err)
+	}
+	return &ResourcesForImport{
+		Pods:            pods,
+		Nodes:           nodes,
+		Pvs:             pvs,
+		Pvcs:            pvcs,
+		StorageClasses:  scs,
+		PriorityClasses: pcs,
+		// existingClusterExportService can't export the SchedulerConfig.
```
sanposhiho, line 47: remove this line

export/utils.go
```diff
@@ -0,0 +1,178 @@
...
+	"golang.org/x/xerrors"
+	corev1 "k8s.io/api/core/v1"
+	schedulingv1 "k8s.io/api/scheduling/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	v1 "k8s.io/client-go/applyconfigurations/core/v1"
+	schedulingcfgv1 "k8s.io/client-go/applyconfigurations/scheduling/v1"
+	cfgstoragev1 "k8s.io/client-go/applyconfigurations/storage/v1"
+)
+
+func ConvertToResourcesForImportFromResourcesForExport(expRes *ResourcesForExport) (*ResourcesForImport, error) {
```
sanposhiho, line 15: Rename it to `ConvertResourcesForImportToResourcesForExport`

sanposhiho (commented):
Looks good overroll, left two more comments.

196Ikuchil:
Thanks!

simulator.go
```diff
@@ -59,6 +68,16 @@ func startSimulator() error {
...
+	// If ExternalImportEnabled is enabled, the simulator import resources
+	// from the existing cluster that indicated by the `KUBECONFIG`.
+	if cfg.ExternalImportEnabled {
+		ctx := context.Background()
+		// This must be called after `StartScheduler`
+		if err := dic.ReplicateExistingClusterService().ImportFromExistingCluster(ctx); err != nil {
+			return xerrors.Errorf("import existing cluster: %w", err)
+		}
+	}
+
```
sanposhiho, line 80: Move this logic to L62 (before starting simulator server)
If the server is started before the import is finished, you will see an empty resource list, depending on the timing of the front-end startup.

sanposhiho (commented):
@196Ikuchil Thanks for this huge work. Approved ❤️ 
(Also thanks @lianghao208 for reviewing this :) )

/lgtm
/approve

PR #112, opened 2022-02-17
Remove `IgnoreErr` options from  the `Import/Export` function called via web API

196Ikuchil:
#### What type of PR is this?
/kind bug

In #103, I have added the `IgnoreErr` option to the `Import/Export` functions called via web API.
But the `IgnoreErr` option was going to use for another logic (Only for Importing the existing Cluster's resources.)
When we call it via web API, it should return an error if an error has occurred.
Since the current usage of the function is incorrect, I fixed it.

And also, I found a typo in the test function's name.

Fixes #103

### DIFF
--- a/export/export_test.go
+++ b/export/export_test.go
@@ -1080,7 +1080,7 @@ func TestService_Import(t *testing.T) {
 	}
 }
 
-func TestService_Import_WithEgnoreErrOption(t *testing.T) {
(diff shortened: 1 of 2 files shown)


196Ikuchil:
Sorry, this is my mistake..

sanposhiho (commented):
Oh, nice catch.
/lgtm
/approve
/ok-to-test

PR #119, opened 2022-02-23
Implement table view on Web UI

### LINKED ISSUE #96: improve Web UI to be able to see many resources for large cluster
In #41, we will import resources from existing cluster. 
Current web UI is hard to see many resources and should be improved.

It would be nice if there was a different mode that would make it easier to see as many resources as possible at once. 
It will probably be simpler than the current UI and will be more like lists of resources. It would also be useful to be able to do things like search for specific resources in the list.

/kind feature

196Ikuchil:
#### What type of PR is this?
/kind feature

- Reorganization of the directories of some current resources List components.
- Implement `DataTable` on web UI. This table view is switchable to the current resources list view.
  - This is an alpha feature. It will be enabled if the `ALPHA_TABLE_VIEWS` environment value is `1`.
- simple searching.
- sorting.
- clickable to edit and delete.

Fixes #96

##### Samples
スクリーンショット 2022-02-28 22 57 05
スクリーンショット 2022-02-28 22 57 16
スクリーンショット 2022-02-28 23 06 15

##### clickable
スクリーンショット 2022-02-28 22 58 12


web/components/ResourceViews/DataTables/NodeDataTable.vue
```diff
@@ -0,0 +1,86 @@
...
+
+    onMounted(getNodeList);
+
+    const nodes = computed(() => nstore.nodes);
+    const onClick = (node: V1Node) => {
+      console.log("aaa");
+      nstore.select(node, false);
+    };
+    const search = "";
+    const headers = [
```
196Ikuchil, line 64: # Need help
We need to discuss which columns should be selected for display.
I think that displaying all parameters is not good.
The current columns of each data table are my randomly selected.

196Ikuchil: In commit [d225fbc](https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/119/commits/d225fbcfdd17243f58278aff835786695d698215), I changed some columns(headers).

These columns are displayed in each table.
- name
- namespace
- capacity
- status(phase)
- condition
- creation-time
- update-time

And for other columns, the unique values of each resource were chosen.

web/components/ResourceViews/DataTables/NodeDataTable.vue
```diff
@@ -0,0 +1,100 @@
...
+      { text: "Status", value: "status.phase", sortable: true },
+      { text: "Capacity", value: "status.capacity", sortable: true },
+      {
+        text: "Creation-Time",
+        value: "metadata.creationTimestamp",
+        sortable: true,
+      },
+      {
+        text: "Update-Time",
+        value: "metadata.managedFields[0].time",
```
196Ikuchil, line 88: currently, we display the `managed field`'s timestamp as an update time.

196Ikuchil: I think that there is no way to get the updated time except this.
I also checked that this doesn't raise an error even if managedFields[0] is nil. (The cell becomes empty.)

sanposhiho:
@196Ikuchil

Some feedbacks. 

For Pod table, 
- `Status` is not needed. (Because it's always pending).

For Node table, 
- `Namespace` is not needed. Node is not namespaced resource. 
- `Conditions ` and `Status` is not needed.
- `Capacity` should show each resource(CPU, memory, and pod)

For PV table, 
- `Namespace` is also not needed. not namespaced resource. 
-  `PVReclaimPolicy` is not needed.

For PVC table, 
- `Status` is also not needed. same reason as Pod.

For SC table, 
- `Namespace` is not needed. same reason.
- `Parameters`, `Reclaim-Policy`, and `Volume-Binding-Mode` are not needed.

For PC table,
- `Namespace` is not needed. same reason.

And for all tables, 
- `Update-Time` is not needed. (Sorry I said that it was needed tho 🙏 )
- The length of the Search box should be the same for all tables.
- change not to use `-`. Use the exact name instead. (like CreationTime, GlobalDefault...)

Others
- Move "List, Table" switch to top bar.

sanposhiho:
And, in current implementation, `Node` on `Pod` table is not updated.
I raise it as another issue because this is not a problem specific to this feature.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/120

196Ikuchil:
Hi @sanposhiho 
Thanks to the reviews. I generally understand.
I have two questions.

1.  `Status `
On `PVC`, there are possibilities to take other values. Is `status` not needed?[Ref:kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#persistentvolumeclaimstatus-v1-core)
(Note: On other resources, I have checked that the `status` parameter is Deprecated. For example, [PodStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#podstatus-v1-core). Therefore, I will remove it from the table.)

2. Move "List, Table" switch to top bar.
I think it shouldn't move. This is because the buttons should not be placed far from the table. Isn't it hard to tell what the switch is for? 
And also the button's component file should be placed under the `ResourceViews` directory.

sanposhiho:
> On PVC, there are possibilities to take other values. Is status not needed?

Yes. Not needed. This value is not much related to scheduling. (This is "scheduler" simulator, so columns that are not related to scheduling are not needed basically.)

> This is because the buttons should not be placed far from the table. Isn't it hard to tell what the switch is for?

It's a fair point. But it is the same even in its current location. It is only when the user touches it that they realize it has something to do with how it is displayed.

How about removing the button and change the UI with env.
Like `ALPHA__TABLE_VIEWS`. If `ALPHA__TABLE_VIEWS` == true (or 1?), use the table to show resources.

sanposhiho:
I think one of the strengths of the current view is "It's easy to see where Pod is."

Conversely, if we can achieve this strength in table view, we can replace the current UI with table view completely.

sanposhiho:
/retitle Implement table view on Web UI

sanposhiho:
> Conversely, if we can achieve this strength in this table view, I think we can replace the current UI with table view completely.

This does not mean that I want you to make changes that way in this PR. I think we can merge this PR first, and then make improvements for that. (Otherwise, this PR will become too big.)

196Ikuchil:
> How about removing the button and change the UI with env.
Like ALPHA__TABLE_VIEWS. If ALPHA__TABLE_VIEWS == true (or 1?), use the table to show resources.

Understood. In this PR, we use `env` value. 
And we expect that the current view will replace this table view in nearly future.

196Ikuchil:
@sanposhiho 
Fixed it✋
I also updated the documentation and screenshots at the top of this PR.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/119#issue-1148249561

web/components/ResourceViews/DataTables/DataTable.vue
```diff
@@ -0,0 +1,59 @@
...
+  },
+  setup() {
+    const search = "";
+
+    return {
+      search,
+    };
+  },
+});
+</script>
```
sanposhiho, line 59:
```suggestion
</script>

<style>
.row-pointer > .v-data-table__wrapper > table > tbody > tr:hover {
  cursor: pointer;
}
</style>

```

web/components/ResourceViews/DataTables/DataTable.vue
```diff
@@ -0,0 +1,59 @@
...
+                label="Search"
+                single-line
+                hide-details
+              ></v-text-field></v-col></v-row></v-card-title
+        ><v-data-table
+          :headers="headers"
+          :items="items"
+          :items-per-page="5"
+          :search="search"
+          multi-sort
```
sanposhiho, line 21:
```suggestion
          multi-sort
          class="row-pointer"
```

sanposhiho (approved):
Thanks @196Ikuchil 👍 
/lgtm
/approve

Issue #46, opened 2021-12-05
support v1 KubeSchedulerConfiguration


sanposhiho:
Currently, the simulator supports v1beta2

https://github.com/kubernetes/enhancements/issues/2891

/kind feature

sanposhiho:
At least, the version needs to be explicitly stated. Currently, users cannot see what's version KubeSchedulerConfiguration is.

![Screen Shot 2021-12-07 at 15 27 59](https://user-images.githubusercontent.com/44139130/144977961-51d7a575-8e26-4d5b-b116-2793e833af2f.png)

markstur:
/assign
I'll look at this next.  I was already showing the kind/apiVersion, but want to look into supporting multiple versions.

sanposhiho:
/remove-lifecycle stale

/triage accepted
Not to mark as stale.

markstur:
Thank you @sanposhiho .  Sorry for the delay. I made progress on this but have been working on other things. Will update soon.

sanposhiho:
No worries :)

sanposhiho:
@markstur Would you mind my taking over this? 
It blocks the version update.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/221

markstur:
Hello @sanposhiho 

I have an update.  I had to rebase/update and figured out that the vendoring, submodule, test updates, etc should be a separate PR before introducing all the config+multipoint changes.  So I just pushed PR and tagged you ( #222 ).  I made that "Part of" this issue, but we should do version updates separately before changing the config and functionality.

Next my multipoint code isn't 100%.  I will take a look at sharing it with you, but I've been blocking this too long so...

YES you can take over. Thank you.

sanposhiho:
Thanks @markstur, I'll review #222 soon. 👍

sanposhiho:
btw, I thought v1beta2 component config was completely removed from k/k at v1.24 😓  (That's why I said this issue is blocking the version update.)

sanposhiho:
So, please feel free to continue to work on this if you like. 
I understood that it's actually not blocking version updates there was no need to rush to work on it.

sanposhiho:
yeah, we should support v1.
/retitle support v1 KubeSchedulerConfiguration

sanposhiho:
memo: #230 needs to be done before it so that we can simply handle plugins registered in multi point.
[...]

PR #123, opened 2022-03-01
display the version of KubeSchedulerConfiguration in WebUI #110

### LINKED ISSUE #110: display the version of KubeSchedulerConfiguration in WebUI
We only support v1beta2 now. The latest version of that is v1beta3, so users may confuse.
So, let's add the version of KubeSchedulerConfiguration in the WebUI.

We plan to support v1beta3 in this issue.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/46

/kind feature
/good-first-issue

markstur:
#### What type of PR is this?

/kind feature

We only support v1beta2 now. The latest version of that is v1beta3, so users may confuse.
So, let's add the version of KubeSchedulerConfiguration in the WebUI.

Fixes #110 


sanposhiho (changes requested):
Hi @markstur, thanks for working on this, and welcome to our project.

So, if there are several ways to do this, or if it is going to be a major change, please discuss it in the issue before doing so. (I should have told this to you when you assigned yourself to the issue. sorry, I didn't notice your `/assign` comment.)

In this PR, there is actually no need to create a new API. I think just adding version information on the scheduler config returned by `GET /schedulerconfiguration` API is enough and a simple way for this. 
And if we proceed this way, it may not need to update the web UI, because the version will be shown in scheduler configuration yaml itself.

If you agree with this, could you update this PR? (or make this change as another PR)

markstur:
Hi @sanposhiho.  It's easy to add the APIVersion to the scheduler configuration shown in the editor, but currently it will let you change it and it won't validate it.  So it's deceptive (e.g., you can set it to beta99 if you want and it sticks).  Also if I try to prefetch the config just to add the version somewhere in the main the UI, then it completely breaks the way the editor/apply works because it currently wants to fetch/edit/post.  If I just pull the version out of whatever is stored, the UI is always one edit behind. Not sure if all that makes sense.

So for this issue... having a hint or label that is very much tied to the go code where the *one* version import happens and showing it near that button seemed best.  It's a hint saying only that version is to be used.  But next...

There's issue #46 to do next. If the first step is to only support v1beta3, then I still like the way this PR handles it. But if we make it accept/validate/show more than one version, then this code could be removed or updated to show the currently used version.  I'll draft a PR for #46 showing version in the yaml and we can take the discussion over there for validate, convert, etc...

(Still trying to figure out where my CLA is stuck.)

sanposhiho:
> it will let you change it and it won't validate it. So it's deceptive (e.g., you can set it to beta99 if you want and it sticks). 

Depends on how we implement it. Just prohibiting (or ignoring) version changes in `POST /schedulerconfiguration` seems enough to solve this problem.
Also, since the version is not easy to see after the gear button is pressed in your design, it is more convenient to be able to see which version while editing yaml. So, displaying the version in yaml, a position that many Kubernetes users are familiar with, makes sense to me.

markstur:
@sanposhiho New PR #129 should be more to your liking.  I did like this one, but it was meant to be temporary anyway.  I think #129 is a better next step and after that I'll look into supporting v1beta3.

We can close this PR if the other one is preferred.  I just used a new PR for now in case folks want to consider both PRs to compare.

/hold

markstur:
Closing in favor of the simpler #129

PR #129, opened 2022-03-03
Display the kind/apiVersion in the scheduler configuration yaml

### LINKED ISSUE #110: display the version of KubeSchedulerConfiguration in WebUI

markstur:
#### What type of PR is this?

/kind feature

We only support v1beta2 now. The latest version of that is v1beta3, so users may confuse.
So, let's add the version of KubeSchedulerConfiguration in the WebUI.

The kind/apiVersion will show in the scheduler configuration yaml when you click on the
setup gear icon.

Only profile edits are supported. The rest are ignored (as documented in the README).
Since prior versions would display yaml edits that were ignored (and this would apply to
the apiVersion now), this PR also ensures that the cleaned versioned yaml with only the profiles
edits is returned.  So if, for example, you try to change the apiVersion and parallelism, it will not
fail on "APPLY", but if you click on setup again you'll see that those changes were discarded.
Support for additional versions is a separate issue #46. A smarter editor and apply validation is also out-of-scope
for this PR, but left for future work.

Fixes #110 


scheduler/scheduler.go
```diff
@@ -102,6 +100,8 @@ func (s *Service) StartScheduler(versionedcfg *v1beta2config.KubeSchedulerConfig
 		return xerrors.Errorf("create scheduler: %w", err)
 	}
 
+	s.currentSchedulerCfg = converted.DeepCopy()
```
sanposhiho, line 103: This `converted` Scheduler Configuration contains the internal plugins for simulator. (because of `convertConfigurationForSimulator`. Described [here](https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/61eebb3e302e8960bcce804e859510c0af71489a/scheduler/scheduler.go#L128).)

So, in this implementation, users can see internal `XXXXXXForSimulator` plugins in SchedulerConfiguration on Web UI. 
スクリーンショット 2022-03-03 21 43 45

From this reason, `s.currentSchedulerCfg` should be updated before `convertConfigurationForSimulator` like previous implementation. 

You changed `convertConfigurationForSimulator` to return versioned config not to allow users to overwrite version, right? So, it may be simpler to overwrite version in [`ApplySchedulerConfig`](https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/61eebb3e302e8960bcce804e859510c0af71489a/server/handler/schedulerconfig.go#L29) instead of modifing version in `convertConfigurationForSimulator`. (yeah, there are other possible solutions to this problem, I'll leave it to you.)

sanposhiho (commented):
Thanks @markstur. I prefer this PR to the previous one.

scheduler/scheduler.go
markstur: @sanposhiho Thank you for catching that and suggesting ApplyConfig.  Yep!  Next commit is much simpler to only apply the profiles and make sure the other changes don't re-occur in the UI.  It really seems to be lacking validation, but I think that will make more sense when I add beta3 for a different PR/issue.  I didn't squash in case you'll want to see both commits, but after it's squashed much of the first will be gone.

sanposhiho:
I will review it this weekend.
/assign

sanposhiho (commented):
Thanks for working on this @markstur!

/lgtm
/approve

PR #134, opened 2022-03-15
Support v1beta3 KubeSchedulerConfiguration

### LINKED ISSUE #46: support v1 KubeSchedulerConfiguration

markstur:
#### What type of PR is this?

/kind feature

Currently the simulator only supports v1beta2 scheduler config. This PR will move it forward to only support the latest v1beta3.

Fixes #46

Issue was not clear if we intend to support both v1beta2 and v1beta3 so this is starting with only v1beta3. If we need both, then we could use this as partial-fix, raise new issue, or just keep working on this PR.


markstur:
/assign @sanposhiho 

I'm assuming you'd like to review.  I feel like the test cases (and my experience) for v1beta3 are kind of lacking.  Advice?

sanposhiho:
@markstur 
Thanks! I'll review it later.

> Issue was not clear if we intend to support both v1beta2 and v1beta3 so this is starting with only v1beta3. If we need both, then we could use this as partial-fix, raise new issue, or just keep working on this PR.

I think just supporting latest version (v1beta3) is enough.

/ok-to-test

sanposhiho:
@markstur 
v1beta3 supports `multipoint` field and we need to consider that. 

https://kubernetes.io/docs/reference/scheduling/config/#multipoint
https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/2891-simplified-config

In current simulator, when users requests to change the scheduler configuration,
- excludes non-allowed changes in configuration.
- replaces filter/score default-plugins with plugins for simulator in [`ConvertForSimulator`](https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/47b354b55b3e2d63c195bd4e904347452fadfc57/scheduler/plugin/plugins.go#L155). 
- replaces filter/score default-plugins config with configs for plugins for simulator in [`NewPluginConfig`.](https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/47b354b55b3e2d63c195bd4e904347452fadfc57/scheduler/plugin/plugins.go#L80)

So, in v1beta3, we need to replace filter/score plugins configured by `multipoint` with plugins for simulator as well.
I expect this will require a slightly more complex change and will be the most difficult part of v1beta3 support.

PR #165, opened 2022-05-08
Implement listwatchresources API on the simulator

### LINKED ISSUE #120: [web ui] Update all resources periodically
update all resources once every 5 seconds.
Users will be able to see all resources without reloading the page even if resources are asynchronously updated. 

/kind feature

196Ikuchil:
#### What type of PR is this?
/kind feature

Implement new API endpoint `/api/v1/listwatchresources` on the simulator.
- This endpoint send all resources's event data to a client by Server-Sent Events.
- We can specify ' lastResourceVersion`.
- If we will not specify it, the endpoint calls `list` API and returns the result before starting `watch`.
- 

Fixes #120 

The implementation of the WebUI side is divided into separate PRs.


196Ikuchil:
This PR includes https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/178.
So, we need to merge it first.
/hold

196Ikuchil:
I have made significant changes to this PR.
We can watch the k8s resources via simulator API.

/unhold

server/server.go
```diff
@@ -46,6 +47,10 @@ func NewSimulatorServer(cfg *config.Config, dic *di.Container) *SimulatorServer
 
 	v1.GET("/export", exportHandler.Export)
 	v1.POST("/import", exportHandler.Import)
+
+	// server push
+	v1.GET("/watchresources", watcherHandler.WatchResources)
```
sanposhiho, line 52: Add this to https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/docs/api.md

docs/api.md
```diff
@@ -99,3 +99,38 @@ You can find sample requests/responses [here](./api-samples/v1/import.md)
 | ----- | -------- |
 | 200   | |
 | 500 | something went wrong (see logs of the simulator server) |
+
+## Watch the simulator's resources
+
+Monitor simulator resources and get updates automatically.
```
sanposhiho, line 105: Need to describe about server-side events so that client can know how to get update.

server/di/di.go
```diff
@@ -120,6 +124,11 @@ func (c *Container) ReplicateExistingClusterService() ReplicateExistingClusterSe
 	return c.replicateExistingClusterService
 }
 
+// WatcherService returns WatcherService.
```
sanposhiho, line 127: hard to understand what this service will do from the name. 
rename to more detailed name. (like `ResourceWatcherService`?)

server/handler/watcher.go
```diff
@@ -0,0 +1,42 @@
...
+	"net/http"
+
+	"github.com/labstack/echo/v4"
+	"k8s.io/klog/v2"
+
+	"github.com/kubernetes-sigs/kube-scheduler-simulator/server/di"
+	"github.com/kubernetes-sigs/kube-scheduler-simulator/watcher"
+)
+
+// WatcherHandler is a handler for watching the simulator's resources.
```
sanposhiho, line 13: nit: 

```suggestion
// WatcherHandler is a handler for watching the k8s resources in the simulator.
```

server/handler/watcher.go
```diff
@@ -0,0 +1,42 @@
...
+type WatcherHandler struct {
+	service di.WatcherService
+}
+
+func NewWatcherHandler(s di.WatcherService) *WatcherHandler {
+	return &WatcherHandler{service: s}
+}
+
+// WatchResources provides a server-pushed response.
```
sanposhiho, line 22:
```suggestion
// WatchResources provides resource updates via server-sent events.
```

watcher/watcher.go
```diff
@@ -0,0 +1,250 @@
...
+// WatchEvent represents an event notified by the watched apiserver.
+type WatchEvent struct {
+	Kind      resourceKind
+	EventType watch.EventType
+	// Obj is an object included in the event notified by the watched apiserver.
+	Obj interface{}
+}
+
+// LastResourceVersions includes each resource's lastResourceVersions.
+// These values are specified from frontend.
```
sanposhiho, line 51: Each service shouldn't know about what a frontend does.

```suggestion
```

watcher/watcher.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	go s.WatchAndHandleEvent(podsEventProxy, podsWatcher, ctx.Done())
+	go s.WatchAndHandleEvent(nodesEventProxy, nodesWatcher, ctx.Done())
+	go s.WatchAndHandleEvent(pvsEventProxy, pvsWatcher, ctx.Done())
+	go s.WatchAndHandleEvent(pvcsEventProxy, pvcsWatcher, ctx.Done())
+	go s.WatchAndHandleEvent(scsEventProxy, scsWatcher, ctx.Done())
+	go s.WatchAndHandleEvent(pcsEventProxy, pcsWatcher, ctx.Done())
+
+	<-ctx.Done()
+	return xerrors.Errorf("terminated to watch: %w", ctx.Err())
```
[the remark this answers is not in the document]
sanposhiho: Nice 👍👍👍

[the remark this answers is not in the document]
sanposhiho: Can you add a comment for this? (A brief summary is enough.)

watcher/watcher.go
```diff
@@ -0,0 +1,250 @@
...
+
+// WatchErrorHandler hanldes some errors.
+func (p *resourceEventProxy) WatchErrorHandler(err error) {
+	switch {
+	case isExpiredError(err):
+		// Don't set LastSyncResourceVersionUnavailable - LIST call with ResourceVersion=RV already
+		// has a semantic that it returns data at least as fresh as provided RV.
+		// So first try to LIST with setting RV to resource version of last observed object.
+		klog.Infof("watch of %v closed with: %w", p.r, err)
+	case xerrors.Is(io.EOF, err):
```
sanposhiho, line 235: Use `errors.Is` instead

sanposhiho (commented):
Rough reviews only for backend side

docs/api.md
```diff
@@ -99,3 +99,38 @@ You can find sample requests/responses [here](./api-samples/v1/import.md)
...
+If you won't specify it, the simulator failed to watch the resource.
+
+|parameter|requirement|description|
+| ----- | --- | -------- |
+|podsLastResourceVersion|MUST|integer|
+|nodesLastResourceVersion|MUST|integer|
+|pvsLastResourceVersion|MUST|integer|
+|pvcsLastResourceVersion|MUST|integer|
+|scsLastResourceVersion|MUST|integer|
+|pcsLastResourceVersion|MUST|integer|
```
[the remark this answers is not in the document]
196Ikuchil: hmm, I don't think so.

Our `/watchresources` API use `NewRetryWatcher ` in internal logic. 
The `NewRetryWatcher` will return error if we pass the "0" to a `lastResourceVersion`. If these values are `optional`, Isn't this likely to be a cause of error?
https://github.com/kubernetes/client-go/blob/3b969f96803febcac81f3a4272a9914270fbb3a3/tools/watch/retrywatcher.go#L66

Also, we shouldn't modify these value.
> Resource versions must be treated as opaque by clients and passed unmodified back to the server.
https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions


Therefore, I think it is generally expected that this `watch API` will be called after the `list API` is called. And at least, I'd liket to require it in this simulator.


BTW, `string` might be better than `integer`.
> You must not assume resource versions are numeric or collatable.

196Ikuchil:
Fixed. (Also some conflicts.)

simulator/server/handler/watcher.go
```diff
@@ -0,0 +1,42 @@
...
+		Scs:   c.FormValue("scsLastResourceVersion"),
+		Pcs:   c.FormValue("pcsLastResourceVersion"),
+	}
+	c.Response().Header().Set(echo.HeaderContentType, echo.MIMEApplicationJSON)
+	c.Response().WriteHeader(http.StatusOK)
+	// Start to watch and do server push
+	err := h.service.WatchResources(ctx, c.Response(), versions)
+	if err != nil {
+		klog.Errorf("closed to watch resources: %+v", err)
+		return echo.NewHTTPError(http.StatusInternalServerError)
```
sanposhiho, line 39: It always returns 500 error when closed. I don't think it's good idea

sanposhiho (commented):
Only server side

docs/api.md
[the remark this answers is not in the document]
sanposhiho: > Our /watchresources API use NewRetryWatcher in internal logic.
The NewRetryWatcher will return error if we pass the "0" to a lastResourceVersion. If these values are optional, Isn't this likely to be a cause of error?

That makes sense.

But, still the API looks unkind to clients. 
What I expect to see in this API is something like `kubectl get Pods -w`. (= After listing/getting the requested object, watch for changes.)

How about deleting these last resource versions params completely? 
In most use cases, the clients need to list all resources directly from kube-apiserver and then call `/watchresources` after that right? 
If so, we don't need to let clients send the last resource versions to server. Just listing all supported objects and start to watch for changes looks more kind to clients.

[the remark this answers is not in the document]
196Ikuchil: > If so, we don't need to let clients send the last resource versions to server. Just listing all supported objects and start to watch for changes looks more kind to clients.

Good question.
Your proposed feature is already provided as this method.
https://pkg.go.dev/k8s.io/client-go/tools/cache#NewInformer

I use `NewRetryWatcher` because it allows me to specify the lastresourceversion at the start of the watch. 

In the future, we will be able to operate several simulators at once.
At that time, it is not desirable to keep WATCHING all simulators.
That is, `watch` needs to stop once, and it needs to be possible to resume `watch` from the middle of the previous one.
And then, In order to restart the watch from the middle, the lastresourceversion needs to be specifiable from the client.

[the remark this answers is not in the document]
196Ikuchil: My new idea was to make the lastresourcecersion optional parameter, and if not specified, to run the `list` on the server side, return the result as an `Added` event to the client, and then start the watch. 

WDYT?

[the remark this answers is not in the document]
196Ikuchil: See also this article.
https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes

simulator/resourcewatcher/resourcewatcher_test.go
```diff
@@ -0,0 +1,484 @@
+package resourcewatcher_test
```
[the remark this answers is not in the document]
sanposhiho: Hm, it doesn't make sense to change test to be in separate package only for solving import cycles.

I'm not sure but seems like `StreamWriter` is one cause...? (Because it uses `WatchEvent` in its definition.)
If so, one idea to solve this is just to create another package for `StreamWriter` and move the related implementation into that new package.

docs/api.md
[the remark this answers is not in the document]
sanposhiho: Ah~, okay. I think now I fully understand your idea.

> My new idea was to make the lastresourcecersion optional parameter, and if not specified, to run the list on the server side, return the result as an Added event to the client, and then start the watch.

After understood your idea, I guess your previous idea, making lastresourcecersion required and using `/watchresources` only to notice the update, makes some sense. 

But, hmmmm, it's a difficult question. I think there are pros/cons for each strategy.

### making lastresourcecersion optional 

This is your new idea. 

**pros**
web UI only need to communicate with `/watchresources` API to fetch the resources. Not need to call kube-apiserver's list API anymore.
That makes the web UI's implementation much simpler, since the client performs the same logic regardless of whether it is the first time access or not. Just accessing `/watchresources` API is always fine.

**cons**
When a user accesses the web UI for the first time, web UI needs to build the resource's view by processing many `Added` event for all resources **one by one**. 
So, If there are too many resources, the building may take some time.

### making lastresourcecersion required 

This is your previous idea.

**pros**

When a user accesses the web UI for the first time, web UI calls list API of kube-apiserver.
By processing another logic for the first time access, the time that takes for first view will be decreased especially when the number of resources is large.

**cons**

Web UI needs to have two logics to show resources; one is for first-time access, and another is for `/watchresources `.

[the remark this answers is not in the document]
sanposhiho: So, to summarize, your new idea vs your previous idea looks like performance vs maintainability.

docs/api.md
```diff
@@ -99,3 +99,45 @@ You can find sample requests/responses [here](api-samples/v1/import.md)
...
+
+
+| parameter                | requirement | description                                                                                        |
+|--------------------------|-------------|----------------------------------------------------------------------------------------------------|
+| podsLastResourceVersion  | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
+| nodesLastResourceVersion | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
+| pvsLastResourceVersion   | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
+| pvcsLastResourceVersion  | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
+| scsLastResourceVersion   | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
+| pcsLastResourceVersion   | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
```
196Ikuchil, line 129: I changed these queries to `OPTIONAL`.

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+}
+
+// ListAndWatch runs list and watch on the target resource. The list is not always ran
+// This method returns error unless an error occurs in the watch. If an error occurs in the watch,
+// it outputs a log and re-run the watch.
+func (s *Service) ListAndWatch(p ResourceEventProxy, stopCh <-chan struct{}) error {
+	lw := createListWatch(p)
+	// If the lastResourceVersion isn't specified by client, call the list and return the result as ADDED event first.
+	if p.LastResourceVersion() == "" {
```
196Ikuchil, line 120: If a client will not set the `lastresourceversion`, the value will be `""`.
And then the server calls `list` method of the resource before starting the `watch`.

web/store/helpers/storeHelper.ts
```diff
@@ -0,0 +1,58 @@
...
+// resourceObject is an interface of stored resource.
+interface resourceObject {
+  metadata?: V1ObjectMeta;
+}
+
+// ResourceState represents a type of each resource's state.
+type ResourceState<T extends resourceObject> = Array<T>;
+
+// createResourceState returns a new resource store's state.
+export function createResourceState<T extends resourceObject>(
```
196Ikuchil, line 12: Almost resources will use this page's methods to handle Events. (Pod is an exception.)
This is because the Pod needs to consider to `unscheduled pod`.

196Ikuchil (commented):
Hi @sanposhiho 
I updated it.

simulator/docs/api.md
```diff
@@ -101,3 +101,45 @@ You can find sample requests/responses [here](api-samples/v1/import.md)
...
+
+We recommend to call the `list` API before the calling and to use `XXXLastResourceVersion` parameters.
+
+The `ResourceVersion` must be treated as opaque by clients and passed unmodified back to the server.
+See also [this page](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions).
+
+
+| parameter                | requirement | description                                                                                        |
+|--------------------------|-------------|----------------------------------------------------------------------------------------------------|
+| podsLastResourceVersion  | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
```
sanposhiho, line 126: Other params as well.

```suggestion
| podsLastResourceVersion  | OPTIONAL    | If not specified, all resources are returned as `ADDED` Events first and then start to watch. |
```

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+// createWatcher creates and returns RetryWatcher.
+func createWatcher(p ResourceEventProxy, lw cache.ListerWatcher) (watch.Interface, error) {
+	rWatcher, err := watchtools.NewRetryWatcher(p.LastResourceVersion(), lw)
+	if err != nil {
+		return nil, xerrors.Errorf("call NewRetryWatcher: %w", err)
+	}
+	return rWatcher, nil
+}
+
+// ResourceEventProxy is an interface that allows handle events and errors.
```
sanposhiho, line 147: move ResourceEventProxy related definitions(func, struct, and interface) to another file in the same package for visibility.

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+type ResourceEventProxy interface {
```
sanposhiho, line 148: `ResourceEventProxy` is used only in this package right? if so, change it to pravate.

sanposhiho: All methods as well.

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+	return func(stopCh <-chan struct{}) error {
+		for {
+			// give the stopCh a chance to stop the loop, even in case of continue statements further down on errors
+			select {
+			case <-stopCh:
+				return nil
+			case event, ok := <-watcher.ResultChan():
+				// grab the event object
+				if !ok {
+					return xerrors.Errorf("closed channel")
```
sanposhiho, line 256:
```suggestion
					return xerrors.New("closed channel")
```

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+				}
+				obj, ok := event.Object.(metav1.Object)
+				if !ok {
+					return xerrors.Errorf("failed to type cast to metav1.Object: %T", event.Object)
```
sanposhiho, line 260:
```suggestion
					return xerrors.Errorf("failed to cast type from %T to metav1.Object", event.Object)
```

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+				case watch.Modified:
+					writingErr = p.writer.Write(&sw.WatchEvent{Kind: p.r, EventType: watch.Modified, Obj: obj})
+				case watch.Deleted:
+					writingErr = p.writer.Write(&sw.WatchEvent{Kind: p.r, EventType: watch.Deleted, Obj: obj})
+				case watch.Bookmark:
+					// A `Bookmark` means watch has synced here, just update the resourceVersion
+				case watch.Error:
+					return xerrors.Errorf("%s: get an error watch event %#v", p.ResourceKind(), obj)
+				default:
+					return xerrors.Errorf("%s: unable to understand watch event: %#v", p.ResourceKind(), obj)
```
sanposhiho, line 275:
```suggestion
					return xerrors.Errorf("%s: unsupported event type %v, object %#v", p.ResourceKind(), event.Type, obj)
```

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+)
+
+// LastResourceVersions includes each resource's lastResourceVersions.
+type LastResourceVersions struct {
+	Pods  string `json:"pods"`
+	Nodes string `json:"nodes"`
+	Pvs   string `json:"pvs"`
+	Pvcs  string `json:"pvcs"`
+	Scs   string `json:"scs"`
+	Pcs   string `json:"pcs"`
```
sanposhiho, line 52: json tags aren't needed?

simulator/docs/api.md
```diff
@@ -101,3 +101,45 @@ You can find sample requests/responses [here](api-samples/v1/import.md)
...
+| nodesLastResourceVersion | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
+| pvsLastResourceVersion   | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
+| pvcsLastResourceVersion  | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
+| scsLastResourceVersion   | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
+| pcsLastResourceVersion   | OPTIONAL    | If not specified, some existing resource data is returned as `ADDED` Events before starting watch. |
+
+e.g.)
+```
+/api/v1/watchresources?podsLastResourceVersion=213&nodesLastResourceVersion=213&pvsLastResourceVersion=213&pvcsLastResourceVersion=213&scsLastResourceVersion=213&pcsLastResourceVersion=213
```
sanposhiho, line 135:
```suggestion
/api/v1/listwatchresources?podsLastResourceVersion=213&nodesLastResourceVersion=213&pvsLastResourceVersion=213&pvcsLastResourceVersion=213&scsLastResourceVersion=213&pcsLastResourceVersion=213
```

simulator/server/server.go
```diff
@@ -46,6 +47,9 @@ func NewSimulatorServer(cfg *config.Config, dic *di.Container) *SimulatorServer
 
 	v1.GET("/export", exportHandler.Export)
 	v1.POST("/import", exportHandler.Import)
+
+	v1.GET("/watchresources", resourcewatcherHandler.WatchResources)
```
sanposhiho, line 51:
```suggestion
	v1.GET("/listwatchresources", resourcewatcherHandler.WatchResources)
```

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+
+// ListAndWatch runs list and watch on the target resource. The list is not always ran
+// This method returns error unless an error occurs in the watch. If an error occurs in the watch,
+// it outputs a log and re-run the watch.
+func (s *Service) ListAndWatch(p ResourceEventProxy, stopCh <-chan struct{}) error {
+	lw := createListWatch(p)
+	// If the lastResourceVersion isn't specified by client, call the list and return the result as ADDED event first.
+	if p.LastResourceVersion() == "" {
+		if err := p.ListAndHandleItems(lw); err != nil {
+			return xerrors.Errorf("call to ListAndHandleItems of %s: %w", p.ResourceKind(), err)
```
sanposhiho, line 122:
```suggestion
			return xerrors.Errorf("call ListAndHandleItems for %s: %w", p.ResourceKind(), err)
```

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+	}
+	watcher, err := createWatcher(p, lw)
+	if err != nil {
+		return xerrors.Errorf("call to createWatcher of %s: %w", p.ResourceKind(), err)
```
sanposhiho, line 127:
```suggestion
		return xerrors.Errorf("call createWatcher for %s: %w", p.ResourceKind(), err)
```

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+	return &resourceEventProxy{
+		writer:              sw,
+		c:                   c,
+		r:                   r,
+		o:                   o,
+		lastResourceVersion: ilrv,
+	}
+}
+
+// ListAndHandleItems calls the list for the resource and the results is send to the client by ListItemsHandler method.
```
sanposhiho, line 187:
```suggestion
// ListAndHandleItems calls the list for the resource, and the results is sent to the client by listItemsHandler method.
```

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+	}
+	var wg wait.Group
+	wg.StartWithChannel(stopCh, run)
+	wg.Wait()
+}
+
+// listItemsHandler sends results of list as "ADDED" event to the client.
+// This method will be expected to call before starting the watch.
+func (p *resourceEventProxy) listItemsHandler(items []runtime.Object) error {
```
sanposhiho, line 234: Let's rename this to sendListedItems. `listItemsHandler ` doesn't tell how `listItems` will be handled by this method.

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+	for _, item := range items {
+		if err := p.writer.Write(&sw.WatchEvent{Kind: p.r, EventType: watch.Added, Obj: item}); err != nil {
+			return xerrors.Errorf("call Write to return list item %#v: %w", item, err)
+		}
+	}
+	return nil
+}
+
+// WatchHandlerFunc watches the specified resource's event and calls the method to send the event
+// and updates lastResourceVersion.
```
sanposhiho, line 244:
```suggestion
// WatchHandlerFunc watches the specified resource's event. 
// When it receives the event from watcher, it sends the event to stream
// and updates lastResourceVersion.
```

simulator/resourcewatcher/resourcewatcher.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+}
+
+func isExpiredError(err error) bool {
+	// In Kubernetes 1.17 and earlier, the api server returns both apierrors.StatusReasonExpired and
+	// apierrors.StatusReasonGone for HTTP 410 (Gone) status code responses. In 1.18 the kube server is more consistent
+	// and always returns apierrors.StatusReasonExpired. For backward compatibility we can only remove the apierrors.IsGone
+	// check when we fully drop support for Kubernetes 1.17 servers from reflectors.
+	return apierrors.IsResourceExpired(err) || apierrors.IsGone(err)
+}
```
sanposhiho, line 323: Seems not needed since our kube-apiserver is v1.22.

sanposhiho (commented):
only for backend

196Ikuchil:
@sanposhiho Thanks! Fixed.

sanposhiho (commented):
@196Ikuchil 
Looks good. Because this PR is so big, could you create the new PR for the frontend and change this PR to only includes the backend changes?
After you do that, I'm gonna approve this.

/lgtm

196Ikuchil:
/retitle Implement listwatchresources API on the simulator

196Ikuchil:
@sanposhiho 
Thank you for the comment.
I move some implementations of webUI to the new PR #211.

PR #172, opened 2022-05-18
Keep priorityclasses shipped with k8s on reset

### LINKED ISSUE #149: Reset function completely deletes all resources, including system's priorityclass
## Problem
The reset function completely deletes all resources, including the system's priorityclass.

After the simulator is launched, these two `PriorityClass`, `system-cluster-critical` and `system-node-critical`, are created automatically.
But this `Reset function` removes these from the simulator, too. 
I think that these two `PriorityClass` should be kept.

↓After the `RESET`.
screenshot

aibarbetta:
#### What type of PR is this?
/kind bug

Use field selector in priorityclasses deletecollection to avoid deleting system priorityclasses

Fixes #149

### DIFF
--- a/priorityclass/priorityclass.go
+++ b/priorityclass/priorityclass.go
@@ -61,6 +61,7 @@ func (s *Service) Delete(ctx context.Context, name string) error {
 
 // DeleteCollection deletes priorityClasses according to the list options.
 func (s *Service) DeleteCollection(ctx context.Context, lopts metav1.ListOptions) error {
+	lopts.FieldSelector = "metadata.name!=system-cluster-critical,metadata.name!=system-node-critical"
 	if err := s.client.SchedulingV1().PriorityClasses().DeleteCollection(ctx, metav1.DeleteOptions{}, lopts); err != nil {
 		return xerrors.Errorf("delete collection of priorityClasses: %w", err)
 	}


sanposhiho:
/assign
/ok-to-test

@aibarbetta Could you add the test case for this?

priorityclass/priorityclass_test.go
```diff
@@ -0,0 +1,135 @@
...
+	t.Parallel()
+	tests := []struct {
+		name                   string
+		prepareFakeClientSetFn func() *fake.Clientset
+		lopt                   metav1.ListOptions
+		wantErr                bool
+		wantReturn             *schedulingv1.PriorityClassList
+	}{
+		{
+			name: "delete all priorityclasses",
```
aibarbetta, line 49: @sanposhiho I attempted to do some testing here, but I believe I cannot actually delete collections with the fake client https://github.com/kubernetes/client-go/issues/609. This test is failing because I still get the deleted priorityclasses after the deletecollection:
```
              	Items: []v1.PriorityClass{
            - 		{ObjectMeta: v1.ObjectMeta{Name: "development"}, Value: 600000},
            - 		{ObjectMeta: v1.ObjectMeta{Name: "production"}, Value: 1000000},
            - 		{ObjectMeta: v1.ObjectMeta{Name: "staging"}, Value: 800000},
              		{ObjectMeta: {Name: "system-cluster-critical"}, Value: 2000000000},
              		{ObjectMeta: {Name: "system-node-critical"}, Value: 2000001000},
              	},
```

sanposhiho: You can use Reactor to check the behavior that isn't supported in fake-client.
https://github.com/kubernetes/client-go/blob/671a40c59fcad6cf1d31592e79dab1effa1768ae/testing/fake.go#L98

aibarbetta:
Hi @sanposhiho, I gave the reactors a try, turns out the fake clientset returned by [`fake.NewSimpleClientset`](https://github.com/kubernetes/client-go/blob/acab036eff5706db06b4eb79253165cf4be0d9c5/kubernetes/fake/clientset_generated.go#L134) doesn't handle the `delete-collection` verb, so I tried adding a reactor for that verb on priorityclasses only. I wanted to avoid creating a new Tracker, so I stuck with the [ObjectTracker interface](https://github.com/kubernetes/client-go/blob/acab036eff5706db06b4eb79253165cf4be0d9c5/testing/fixture.go#L43), I used `tracker.List` and `tracker.Delete` to build a `"tracker.DeleteCollection"` functionality. Tests are fixed now, but let me know if there is a nicer way to achieve this.

PR #178, opened 2022-05-22
apiserver supports TLS

### LINKED ISSUE #175: TLS supports on `apiserver`
Let's support the `TLS` connection in the development environment.

We'd like to use multiple stream connections to apiserver via `http2`. (See also #120)
However, if we use `http2', we must use `HTTPS` for the connection from browser to the `apiserver`.

196Ikuchil:
#### What type of PR is this?
/kind feature

In #120, we need to support TLS connection on apiserver.
Therefore, this PR allows us to do two things
- Use `HTTP/2` to communicate with apiserver. (This required the use of TLS.)
- Set and use our own certificate file.

Fixes #175

References
- https://pkg.go.dev/net/http/internal/testcert#pkg-variables
- https://docs.vmware.com/en/VMware-Adapter-for-SAP-Landscape-Management/2.1.0/Installation-and-Administration-Guide-for-VLA-Administrators/GUID-D60F08AD-6E54-4959-A272-458D08B8B038.html
- https://unix.stackexchange.com/questions/644176/how-to-permanently-add-self-signed-certificate-in-firefox
- https://stackoverflow.com/questions/54899550

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,14 @@ It can be used to learn about the Kubernetes scheduler or to examine the detaile
 
 ## Getting started
 
+### Prepare
+You should import apiserver's certificate on your browser,
+since the simulator frontend uses `HTTP/2` connections for some communications.
+
+See [README.md](./web/README.md#importing-certificate) for more information on importing certificate.
+And if you would like to use your own certificate, see also [this section](./docs/kube-apiserver.md#environment-variable)
+
+
 ### Run with Docker
 
 We have [docker-compose.yml](./docker-compose.yml) to use the simulator easily.
--- a/config/config.go
+++ b/config/config.go
@@ -74,7 +74,7 @@ func NewConfig() (*Config, error) {
 		Port:                  port,
 		KubeAPIServerURL:      apiurl,
 		KubeAPIServerCertPath: apicertpath,
-		KubeAPIServerKeyPath:	 apikeypath,
+		KubeAPIServerKeyPath:  apikeypath,
 		EtcdURL:               etcdurl,
 		CorsAllowedOriginList: corsAllowedOriginList,
 		InitialSchedulerCfg:   initialschedulerCfg,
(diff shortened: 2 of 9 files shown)


196Ikuchil:
This change makes the development environment a little more difficult for developers.
And also, we will need to verify that this change is actually required for #120.

/hold

196Ikuchil:
We verified that this PR is needed for https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/165.
So, please review this?

/unhold

sanposhiho:
Hmmmm, the challenge itself is really appreciated, but I don't like this requires the additional manual setting to use web UI... 😓 
If we cannot achieve this way with a completely automated way (something like `make generate-and-set-cert`), how about creating the API in the simulator server to watch all resource and using Web Sockets between the simulator server and web UI? (Does the web socket have the same issue of the limitation of connections?)

i.e. Web UI ←(web socket)→ the simulator server ←(SSE)→ kube-apiserver
1. Web UI access the endpoint (named like `/api/v1/watch`?) when users access.
2. The simulator starts to watch changes in all supported resources.
3. The simulator sends the message to WebUI when some resource is changed.

196Ikuchil:
@sanposhiho 
We would not require the manual setting if used a verified certificate (not a self-signed certificate).
But I agree with you also.

> i.e. Web UI ←(web socket)→ the simulator server ←(SSE)→ kube-apiserver

It seems strange to me that there are two ways to get and update resources.

> (Does the web socket have the same issue of the limitation of connections?)

I think it maybe no problem. And if we call only the way (`/api/v1/watch`), the limitation of the `stream` is not a problem, too.

196Ikuchil:
I discussed this with @sanposhiho in DM.
As a result, we aim to implement like this

Web UI ←(stream)→ the simulator server ←(watch API)→ kube-apiserver

In the above implementation, we don't need TLS.(Because we won't use `HTTP/2`)
Therefore, this PR may not be required, currently.

/hold

196Ikuchil:
Currently, we will not need to TLS for #120.

/close

Issue #167, opened 2022-05-12
New feature Idea: scenario-based benckmark testing on kube-scheduler-simulator


sanposhiho:
https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/140 tries to add the scenario based simulation feature to kube-scheduler-simulator and it will be one of the usecases of kube-scheduler-simulator-operator.
(For more details about the scenario based simulation, please see https://bit.ly/scenario-based-scheduler-simulation.)

With the similar way, we can support benchmark testing for scheduler. 

I don't know yet if this should be defined as a new CRD or if it should be supported by extending Scenario CRD. I will mention those options when I write the proposal for this.

/kind feature
/assign
/priority important-longterm

196Ikuchil:
I'm going to give it a shot👌
/assign

sanposhiho:
/remove-area simulator
Sorry, we can't know where it will be implemented yet.

sanposhiho:
We've got a KEP for this already. I'll close this and will open other issues for the development of this feature.

/close

Issue #159, opened 2022-04-30
New feature idea: kube-scheduler-simulator-operator


sanposhiho:
## What I want to introduce

custom resource `Simulator` and custom controller to manage that

---

The controller will be included in the simulator backend binary/container image. 
And what I'd like to propose is not create `Simulator` CRD on real kube-apiserver, but create it on kube-apiserver included in our simulator. (This could be a point for discussion tho.) All users need to do to use kube-scheduler-simulator-operator is only deploy kube-scheduler-simulator in their cluster with the permission to create Pods on their cluster. (So, they don't need to deploy controller or create CRD by themselves!)

Yeah, this feature requires that the simulator is deployed on Kubernetes since the controller need to start simulators as new container.
I believe this feature brings new and great possibilities to this project.

## Why I want to introduce it

With this, we can easily create/use the simulator as sandbox for the simulation. This feature makes it easier to handle multiple simulators and to create or modify simulators from the outside via kubectl or any other kube-apiserver clients.

#140 wants to add the scenario based simulation feature to kube-scheduler-simulator and it will be one of the usecases of kube-scheduler-simulator-operator. 
For more details about the scenario based simulation, please see https://bit.ly/scenario-based-scheduler-simulation.

We might add `SimulatorName` to Scenario's spec and the Scenario will be run on that simulator. One big advantage of this is we can use a different scheduler without re-deploying the simulator. (Currently, scenarios, created in the simulator, will run on the same simulator and it means all scenarios will be run on a same simulator.)

I'll write the detailed proposal later.

/kind feature
/assign
/priority important-longterm

matthewygf:
+1 for the SimulatorName for each different scenarios!

sanposhiho:
/triage accepted
Not to mark as stale.

sanposhiho:
We've got a KEP for this already. I'll close this and will open other issues for the development of this feature.

/close

PR #185, opened 2022-06-13
Add KEP-167 'scheduler simulator operator'

sanposhiho:
#### What type of PR is this?

/kind feature
/kind documentation

Add KEP-167 'scheduler simulator operator'

Part of #167


196Ikuchil:
@sanposhiho 
Thank you for your awesome proposal.
I'd like to confirm, but is this PR for #167? I think that this is for #159.
Please check it.

sanposhiho:
Ahhh, you are right. Let me fix it later. 👍

keps/159-scheduler-simulator-operator/README.md
```diff
@@ -0,0 +1,118 @@
+# KEP-159: scheduler simulator operator
+
+## Summary
+
+A new custom resource to represent a simulator itself and a custom controller to manage it are introduced.
```
196Ikuchil, line 5:
```suggestion
A new custom resource to represent a simulator itself and a custom controller to manage it will be introduced.
```
?

keps/159-scheduler-simulator-operator/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Proposal
+
+### Implementation design details
+
+#### Required prior knowledge
+
+The current simulator is built in a single binary: the binary contains kube-apiserver, kube-scheduler and the simulator backend (which provides the web UI specific feature.)
+
+And, the simulator can be configured via environment variables
```
196Ikuchil, line 31:
```suggestion
And, the simulator can be configured via environment variables.
```

keps/159-scheduler-simulator-operator/README.md
```diff
@@ -0,0 +1,118 @@
...
+  Status SimulatorStatus
+}
+
+// SimulatorSpec has the spec for the simulator.
+// All fields can be updated (but a Pod may be re-created.)
+type SimulatorSpec struct {
+  // KubeAPIServerPort indicates kube-apiserver’s URL which the simulator has internally.
+  // (The each simulator has own kube-apiserver internally.)
+  // This field's value will be added to a "simulator" container's env as "KUBE_API_PORT".
+  KubeAPIServerPort            int
```
196Ikuchil, line 54: What is the `KubeAPIServerPort ` indicate? Port(int) or URL(string)?

196Ikuchil (commented):
I will re-review later✋

keps/159-scheduler-simulator-operator/README.md
```diff
@@ -0,0 +1,118 @@
+# KEP-159: scheduler simulator operator
```
everpeace, line 1:
```suggestion
# KEP-159: Scheduler Simulator Operator
```

keps/159-scheduler-simulator-operator/README.md
```diff
@@ -0,0 +1,118 @@
...
+## Motivation
+
+With this, we can easily create/use the simulator as sandbox for the simulator by creating the Simulator custom resource. 
+
+This feature also makes it easier to handle simulators from other CRD or controllers. 
+They can gets the information for the simulator easily by accessint the `Simulator` resource.
+
+### Goals
+
+- Users can define simulators with Simulator CRD.
```
everpeace, line 16:
```suggestion
- Users can define simulators via `Simulator` CRs.
```

keps/159-scheduler-simulator-operator/README.md
```diff
@@ -0,0 +1,118 @@
...
+type Simulator struct {
+  metav1.TypeMeta 
+  metav1.ObjectMeta // namespaced.
+
+  Spec SimulatorSpec
+  Status SimulatorStatus
+}
+
+// SimulatorSpec has the spec for the simulator.
+// All fields can be updated (but a Pod may be re-created.)
```
everpeace, line 49: Don't we need to handle existing(finished/running/pending) `Scenario` resources inside of the sumulator??  I think we should wait for users to clean up `Scenario` in the simulator. What do you think?

keps/159-scheduler-simulator-operator/README.md
```diff
@@ -0,0 +1,118 @@
...
+  PodTemplate *corev1.PodTemplateSpec 
+}
+
+type SimulatorStatus struct {
+  Phase SimulatorPhase
+
+  // A human readable message indicating details about why the simulator is in this phase.
+  // optional
+  Message *string 
+}
```
everpeace, line 101: I think it would be useful to kube-apiserver's url, kubeconfig, etc. information in the status.  Users will use them to create `Scenario` CRs in the simulator.

everpeace (commented):
Thanks for the write-up! Please see the comments and suggestions.

keps/159-scheduler-simulator-operator/README.md
sanposhiho: It's Port.

keps/159-scheduler-simulator-operator/README.md
sanposhiho: Good question.

> we should wait for users to clean up Scenario in the simulator.

What did you mean by "clean up"? Let users delete all Scenario resources in the simulator? So, I don't want to let users do something for restarting the simulator Pod (since it's hard for users to realize they need to do that to kick restarting), but want the `Simulator` controller to do all the things for restarting the simulator Pod. 

How about this:
- implementing a way (endpoint on the simulator server?) that tells the Scenario controller that the Simulator controller wants to shut down the Pod.
  - when the Scenario controller moves FakeTime forward, it checks if the shutdown notification has come or not. If it has, the Scenario controller changes the Scenario's phase to "Paused" instead of continuing running the Scenario.
- setting `preStop` to notify shutdown to the Scenario controller by the above way.

sanposhiho: I think we can put this in "Future work", and until implementing them, the controller will behave like this: when the controller is started and finds the Scenario which phase is "Running", the controller just changes the status "Failed" with the `.status.message` like "the controller restarted while the Scenario was running".

keps/159-scheduler-simulator-operator/README.md
sanposhiho: sg 👍

keps/159-scheduler-simulator-operator/README.md
everpeace: > Let users delete all Scenario resources in the simulator?

Yeah, I thought so.

> How about this:

that sounds reasonable about handling the running Scenario. How about completed scenarios? If a Simulator CR (i.e. simulator pod) hosts multiple (say a lot of) completed Scenario CRs and what if a user updated or deleted the Simulator CR accidentally??  I think we should let users resurrect Scenario results in some way.

sanposhiho: >  How about completed scenarios?

We don't need to do anything. The etcd has the data for completed scenario.

> what if a user updated or deleted the Simulator CR accidentally??

Could you elaborate on that? The system should not resurrect what our users have updated/deleted.

keps/159-scheduler-simulator-operator/README.md
sanposhiho: Ooh, okay

keps/159-scheduler-simulator-operator/README.md
sanposhiho: > I think we can put this in "Future work"

I'll write this in https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/186

sanposhiho: Added the note for this on https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/186

keps/159-scheduler-simulator-operator/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+type SimulatorStatus struct {
+  Phase SimulatorPhase
+
+  // A human readable message indicating details about why the simulator is in this phase.
+  // optional
+  Message *string 
+
+  // KubeAPIServerURL is the url to access the kube-apiserver in the simulator.
+  KubeAPIServerURL string
```
everpeace, line 106: Let's add `KubeConfig` or `KubeConfigPath`.

everpeace (changes requested):
LGTM overall. just one comment.

keps/159-scheduler-simulator-operator/README.md
```diff
@@ -0,0 +1,126 @@
...
+  // - have environment variables to configure the simulator.
+  // 
+  // And if the EtcdURL field is empty, the system will add the container named "etcd" for kube-apiserver.
+  // 
+  // You can overwrite those container spec by defining containers that have the same name, "simulator" or "etcd".
+  // 
+  // When PodTemplate has the container(s) named "simulator" or "etcd" and the field is different from the container spec generated by other fields in SimulatorSpec,
+  // container(s) definition in PodTemplate takes precedence.
+  PodTemplate *corev1.PodTemplateSpec 
+}
```
196Ikuchil, line 96: What do you think about the `EXTERNAL_IMPORT_ENABLED` env for the external cluster import function?
Currently, the function is Beta one, and it will not be required to `Scenario-based simulation` and `scheduler-simulation`. But I think that it is important a part of the essential function for future.

196Ikuchil (commented):
@sanposhiho 
Thanks for the update!
I left a new comment.

keps/159-scheduler-simulator-operator/README.md
sanposhiho: I think it's not required at least for now since users can set EXTERNAL_IMPORT_ENABLED in PodTemplate field if they want to use.

keps/159-scheduler-simulator-operator/README.md
sanposhiho: As we discussed offline, KubeConfig isn't needed since users can communicate with the kube-apiserver in simulator only with url.

196Ikuchil:
Thanks!

/lgtm

PR #186, opened 2022-06-13
Add KEP-140 'Scenario-based simulation'

sanposhiho:
#### What type of PR is this?

/kind feature
/kind documentation

#### What this PR does / why we need it:

Add KEP-140 'Scenario-based simulation'

#### Which issue(s) this PR fixes:

Part of #140

#### Special notes for your reviewer:

/label tide/merge-method-squash

Issue #184, opened 2022-06-12
`SchedulerSimulation` CRD to create `Simulator` and run the `Scenario` in the created Simulator


sanposhiho:
New `SchedulerSimulation` CRD and controller for it.

The `SchedulerSimulation` resource represents one-shot scheduler simulation, and by creating `SchedulerSimulation` resource in cluster, users can run `Simulator` and run `Scenario` in that simulator.
(`Scenario` is the CRD to run scenario-based simulation introduced in https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/140.)
(`Simulator` is the CRD to run the Pod for simulator introduced in https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/159)

/assign
/kind feature
/triage accepted
/priority important-longterm

PR #187, opened 2022-06-13
Add KEP-184 'scheduler simulation'

sanposhiho:
#### What type of PR is this?

/kind feature
/kind documentation

Add KEP-184 'scheduler simulation'

Part of #184

This feature is built over the features introduced in KEP-140 and KEP-167. So, this PR needs to be merged after them.

Note that this KEP may need to be significantly changed if there are significant changes from what is currently proposed in KEP-140 and KEP-167.

https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/185
https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/186

### DIFF
--- a/keps/184-scheduler-simulation/README.md
+++ b/keps/184-scheduler-simulation/README.md
@@ -73,21 +73,14 @@ type SchedulerSimulationSpec struct {
   //
   // Cannot be updated.
   ScenarioResultFilePath string
-  // Volume to mount into the filesystem.
+  // Volume to be mounted.
   //
   // The container named "scenario-runner" will be created in the simulator Pod.
   // That container will see the scenario defined in ScenarioTemplateFilePath, create the scenario in the simulator, and store the result into ScenarioReusltFilePath.
-  // The volume specified in Volume and VolumeMount is mounted in that container.
+  // The volume specified here is mounted as mountPath `/mnt` in that container.
   //
-  // Cannot be updated.
-  VolumeMount corev1.VolumeMount 
-  // Volume that can be mounted by VolumeMount field.
+  // More info about volume: https://kubernetes.io/docs/concepts/storage/volumes
   //
-  // The container named "scenario-runner" will be created in the simulator Pod.
-  // That container will see the scenario defined in ScenarioTemplateFilePath, create the scenario in the simulator, and store the result into ScenarioReusltFilePath.
-  // The volume specified in Volume and VolumeMount is mounted in that container.
-  //
-  // More info: https://kubernetes.io/docs/concepts/storage/volumes
   // Cannot be updated.
   Volume corev1.Volume 
   // ScenarioRunnerImage is the image for the container named "scenario-runner" that will be created in the simulator Pod.
--- a/keps/184-scheduler-simulation/kep.yaml
+++ b/keps/184-scheduler-simulation/kep.yaml
@@ -18,4 +18,4 @@ milestone:
 
 see-also:
   - "/keps/140-scenario-based-simulation"
-  - "/keps/167-scheduler-simulator-operator"
+  - "/keps/159-scheduler-simulator-operator"


sanposhiho:
> This feature is built over the features introduced in KEP-140 and KEP-167. So, this PR needs to be merged after them.
Note that this KEP may need to be significantly changed if there are significant changes from what is currently proposed in KEP-140 and KEP-167.

To avoid accidental merging. 

/hold

keps/184-scheduler-simulation/README.md
```diff
@@ -0,0 +1,214 @@
...
+  // Cannot be updated.
+  ScenarioResultFilePath string
+  // Volume to mount into the filesystem.
+  //
+  // The container named "scenario-runner" will be created in the simulator Pod.
+  // That container will see the scenario defined in ScenarioTemplateFilePath, create the scenario in the simulator, and store the result into ScenarioReusltFilePath.
+  // The volume specified in Volume and VolumeMount is mounted in that container.
+  //
+  // Cannot be updated.
+  VolumeMount corev1.VolumeMount
```
everpeace, line 83: Do we need to allow users to configure the path/volumeMount for them?? I suppose the fixed path/volumeMount would be enough, i.e., this raises any inconvenience for them. WDYT?

everpeace (commented):
Thanks for writing up!!  I left a few comments.  This looks good to me.  

Although I think the kep stays looks good to me when many changes happened in dependents keps, let's see what's going on them　for now.

keps/184-scheduler-simulation/README.md
sanposhiho: > I suppose the fixed path/volumeMount would be enough, i.e., this raises any inconvenience for them. WDYT?

I may not have understood exactly what you meant. So, you wanna remove `VolumeMount` and make `Volume` always be mounted in the same `mountPath` in "scenario-runner" container?

keps/184-scheduler-simulation/kep.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - "@196Ikuchil"
+
+stage: alpha
+
+milestone:
+  alpha: "v0.1"
+
+see-also:
+  - "/keps/140-scenario-based-simulation"
+  - "/keps/167-scheduler-simulator-operator"
```
196Ikuchil, line 21:
```suggestion
  - "/keps/159-scheduler-simulator-operator"
```

keps/184-scheduler-simulation/README.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+#### SchedulerSimulation execution flow
+
+0. User creates SchedulerSimulation resource in user's cluster.
+1. The SchedulerSimulation controller creates a Simulator resource with "scenario-runner" container.
+2. After the Pod gets started, the "scenario-runner" reads ScenarioTemplateFile and creates the Scenario resource in simulator's kube-apiserver.
+3. The simulator run the scenario.
+4. After the scenario is done, the "scenario-runner" fetch the Scenario's result from simulator's kube-apiserver and store it in ScenarioResultFile.
+5. The "scenario-runner" container is terminated.
```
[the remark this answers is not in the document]
sanposhiho: Nice question. All your guessing is correct.

Users can define the scenario that doesn't have "Done" operation. And, in that case, the user can access the simulator to add the more scenarios by looking at the simulation results and the status of the resource at that point.

keps/184-scheduler-simulation/kep.yaml
sanposhiho: nice catch. will fix later

keps/184-scheduler-simulation/README.md
sanposhiho: make sense.

sanposhiho:
@everpeace @196Ikuchil Updated as your reviews.

everpeace (approved):
LGTM!

*I put off `/lgtm` not to get this PR merged because other reviewers are still in review.*

196Ikuchil:
Thanks!
This feature would make the simulator much more useful!

/lgtm

Issue #182, opened 2022-06-03
Support external scheduler


sanposhiho:
There is great value for this project to support many types of schedulers.

Just replacing [here](https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/9de8c472f348b31437cce5ca2a34506f874cdddb/scheduler/scheduler.go#L86) to create user's scheduler is the easy way for users to use their scheduler in simulator, but, if user's scheduler is based on different version of Kubernetes or built on a completely different mechanism than kube-scheduler in k/k, it doesn't work.

For those schedulers, I propose here to support "external scheduler". 

Users can run any type of scheduler as outside schedulers and that scheduler will communicate with kube-apiserver in the simulator to run simulation on the simulator.

There is drawback to outside scheduler: they cannot change scheduler configuration via Web UI (or simulator API).
But, the big advantage of this is users can use their scheduler with the other advanced features of the simulator like Scenario.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/140

/kind feature
/triage accepted
/priority important-longterm

sanposhiho:
give this a higher priority. It will important both for version issue and scenario API.

sanposhiho:
/remove-priority next-release

Still quite important, but just remove it from next release candidate.

hzliangbin:
@sanposhiho how to use external scheduler? is there a doc?

sanposhiho:
We're now working on [the prioritized issues](https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues?q=is%3Aissue+is%3Aopen+label%3Apriority%2Fnext-release) for the first release.
We'll work on this after that. (so, not immediately)

hzliangbin:
@sanposhiho it's ok. just want an instruction for config external scheduler.

PR #194, opened 2022-07-07
Changed the design to allow Scenario to be used with more kind of controllers

sanposhiho:
#### What type of PR is this?

/kind documentation

Propose the way to stop scheduling via ScenarioStatus.SchedulerStatus

---

Add `SchedulerStatus` on Status which indicates the scheduler's current status.
Both scheduler and the scenario controller will change this field, see more detail in the change of this PR.

By going this way, we don't need to create the scenario controller in simulator; we can create it as an isolated binary.
That means the scenario controller can be used without simulator. (Like in minikube, kind or anything else) and Scenario can be used in the outside scheduler as well. (https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/182)

Fixes #

/assign @196Ikuchil 
/cc @everpeace 


keps/140-scenario-based-simulation/README.md
```diff
@@ -148,6 +151,25 @@ type ScenarioStatus struct {
   ScenarioResult ScenarioResult
 }
 
+
+type SchedulerStatus string
+const (
+  // SchedulerWillRun indicates the scheduler is expected to start to schedule.
+  // In other words, the scheduler is currently stopped, 
+  // and will start to schedule Pods when the state is SchedulerWillRun.
+  SchedulerWillRun  SchedulerStatus = "WillRun"
```
sanposhiho, line 160: Are `WillRun`, "WillStop" weird name? 😅 
Let me know if there are other good names for them

keps/140-scenario-based-simulation/README.md
```diff
@@ -298,10 +320,13 @@ ScenarioStep is:
...
 
 1. run all operations defined for that step.
-2. scheduler starts scheduling.
-3. scheduler stops scheduling when it can no longer schedule any more Pods.
-4. update status.scenarioResult.
-5. move to next step.
+2. the scenario controller changes status.SchedulerStatus to SchedulerWillRun.
+3. the scheduler starts scheduling and changes status.SchedulerStatus to SchedulerRunning.
+4. the scenario controller changes status.SchedulerStatus to SchedulerWillStop, 
+when it can no longer schedule any more Pods.
```
everpeace, line 326: In general, I think it would be impossible for the simulator to detect the scheduler finishes the scheduling in this step.  So, we add a note for the assumptions for schedulers that the simulator support.  I think another PR is ok.

keps/140-scenario-based-simulation/README.md
everpeace: how about introducing two fields that are _simulator expects to_ and _reported from scheduelr_.

```yaml
spec:
  # expected scheduler status by the simulator
  schedulerStatus: "Running|Stopped"
status:
  # reported status from the scheduler
  schedulerStatus: "Running|Stopped"
```

everpeace (changes requested):
I put several comments. And, shall we add a note to `ScenarioPodScheduleResult` that `AllCandidateNodes,AllFilteredNodes,PluginResult` will be all blank if the scheduler is NOT scheduling framework based one?

keps/140-scenario-based-simulation/README.md
sanposhiho: That's another idea. Actually I considered that idea when I wrote this, but, it was rejected in my head. 
What I thought about it at that time was:
- If we add an expected scheduler status in ScenarioSpec like what you said, literary that means `expected scheduler status` is part of the spec of "scenario". 
   - It's weird, isn't it? The weirdness comes from the fact that "the expected scheduler's status actually depends on the **status** of the Scenario.".
   - In other words, from the user's point of view, they will never need to touch that field since it's automatically changed by the controller depending on the stage of one `ScenarioStep`.
- I guess if we want to really put it on k8s standard way, we ideally need to create another CRD like `Scheduler` or `SimulationScheduler`. But that's too much. (at least for now.)
    - And, even if we consider creating the `Scheduler` CRD in the future, we can easily migrate the Scenario's `status.SchedulerStatus`; we can keep `status.SchedulerStatus` in that case.

That's why I prefer the current `.status.SchedulerStatus` strategy.

keps/140-scenario-based-simulation/README.md
sanposhiho: > In general, I think it would be impossible for the simulator to detect the scheduler finishes the scheduling in this step. 

You mean we cannot support all schedulers in the world because the possibilities for how to implement a scheduler are endless, right?
I added the answer at bottom of the KEP in this PR. ("can scenario work with all kind of schedulers?" section)
Please check it.

sanposhiho:
> shall we add a note to ScenarioPodScheduleResult that AllCandidateNodes,AllFilteredNodes,PluginResult will be all blank if the scheduler is NOT scheduling framework based one?

👍 
will do it later.

keps/140-scenario-based-simulation/README.md
```diff
@@ -47,246 +47,307 @@ We may need to change etcd request-size limitation by --max-request-bytes since
...
+	PreemptedAt *ScenarioStep `json:"preemptedAt"`
+	// ScheduleResult has the results of all scheduling for the Pod.
+	//
+	// If the scheduler working with a simulator isn't worked on scheduling framework,
+	// this field will be empty.
+	// TODO: add the link to doc when it's empty.
+	//
+	// +patchStrategy=replace
+	// +optional
+	ScheduleResult []ScenarioPodScheduleResult `json:"scheduleResult"`
```
everpeace, line 316: Thanks 👍

everpeace (approved):
LGTM!

keps/140-scenario-based-simulation/README.md
```diff
@@ -47,246 +47,307 @@ We may need to change etcd request-size limitation by --max-request-bytes since
...
+	// SchedulerWillRun indicates the scheduler is expected to start to schedule.
+	// In other words, the scheduler is currently stopped,
+	// and will start to schedule Pods when the state is SchedulerWillRun.
+	SchedulerWillRun SchedulerStatus = "WillRun"
+	// SchedulerRunning indicates the scheduler is scheduling Pods.
+	SchedulerRunning SchedulerStatus = "Running"
+	// SchedulerWillStop indicates the scheduler is expected to stop scheduling.
+	// In other words, the scheduler is currently scheduling Pods,
+	// and will stop scheduling when the state is SchedulerWillStop.
+	SchedulerWillStop SchedulerStatus = "WillStop"
```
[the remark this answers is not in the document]
sanposhiho: > The period of time that the status is SchedulerWillStop is very short?

Most cases, yes.
`SchedulerWillStop`'s purpose is, in other words, asking the scheduler to stop scheduling.
So, when (4), the scenario controller wants scheduler to stop scheduling, and change SchedulerStatus to `SchedulerWillStop`. And the scheduler will notice the scenario controller wants to stop scheduling and stop it. (at (5)

Does it make sense?

sanposhiho:
@196Ikuchil @everpeace 
Updated in https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/194/commits/87017e9b121326fbf41b4d22fc73f3c2da3a9e9d.

Please take a look again when have a chance.

- create `StepPhase` and use it instead of `schedulerStatus`.
  - The names are changed, but the purpose behind the field is the same.
- judge it can no longer schedule any more Pods in scheduler's `NextPod`.
- specify the supported scheduler

sanposhiho:
/retitle elaborate more on how to/when stop scheduling

everpeace (changes requested):
I left some comments. And,

- I see several grammatical issues.  Please proofread again.
- I would like to propose "external" scheduler instead of "outside."  What do you think?

I would add more comments perhaps.

keps/140-scenario-based-simulation/README.md
```diff
@@ -47,276 +47,421 @@ We may need to change etcd request-size limitation by --max-request-bytes since
...
 
+#### supported scheduler in simulator
+
+The goal of this KEP is to make schedulers that meet the all following criteria work with scenario:
+
+1. schedule Pods by the conditions of k8s resources.
+    - In other words, schedulers shouldn't schedule Pods by the conditions other than k8s resources.
+    - For example, if the scheduler that schedules Pods by checking some metrics from external server, 
+    the scenario controller cannot manage metrics value and the Scenario may yield different results for each run.
+2. scheduler that doesn't creates/edits/deletes resource. (except preemption and binding)
```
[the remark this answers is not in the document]
sanposhiho: > I think we should elaborate on this constraint more accurately. Technically, kube-scheduler updates pod's status.nominatedNodeName and creates Event resources.

Definitely. will update that.

sanposhiho:
/retitle Changed the design to allow Scenario to be used with more kind of controllers

sanposhiho:
@everpeace @196Ikuchil 

Applied a big update.... 🙇 🙏  Please take a look again.

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,316 +64,588 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
+All controllers that refer only to k8s resources won't do anything if there is no change on k8s resources state.
+In other words, all controllers may do something only when someone performs a k8s resource operation.
+
+#### Expect when controllers finish working
+
+In the simulator, users define the `WaitForControllerFinish` function for all controllers so that we can expect when a controller finish working.
+
+```go
+// ControllerBehaviorDefinition has the method to define a behavior for a controllers.
+type ControllerBehaviorDefinition interface {
```
sanposhiho, line 445: Not so important here, but looking for a better name for this interface 😅 .

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,65 +64,85 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
-	// ID for this event. Normally, the system sets this field for you.
+type ControllerSet struct {
+	// Enabled specifies controllers that should be enabled.
+	// +listType=atomic
+	Enabled  []Controller `json:"enabled"`
+	// Disabled specifies controllers that should be disabled.
+	// When all controllers need to be disabled, an array containing only one "*" should be provided.
+	// +listType=map
+	// +listMapKey=name
+	Disabled []Controller `json:"disabled"`
```
196Ikuchil, line 109: What is this mean if we register a controller to `Disabled` of SimulatedControllers. Is there some effect?

196Ikuchil (commented):
quick check 💨

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,65 +64,85 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
-	DoneOperation *DoneOperation `json:"doneOperation,omitempty"`
+	Done *DoneOperation `json:"doneOperation,omitempty"`
 }
 
 // OperationType describes Operation.
-// Please see the following defined OperationType, all operation types not listed below are invalid.
+// Please see the following defined OperationType; all operation types not listed below are invalid.
 type OperationType string
 
 const (
```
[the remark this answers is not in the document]
sanposhiho: Ah~, yes. I think we can remove them.

keps/140-scenario-based-simulation/README.md
```diff
@@ -174,58 +215,54 @@ type ScenarioStatus struct {
...
+	// OperatingCompleted means the preparing controllers have finished operating operation defined for the step.
+	OperatingCompleted StepPhase = "OperatingCompleted"
+    // ControllerRunning means the simulated controller is working.
+    ControllerRunning StepPhase = "ControllerRunning"
+    // ControllerPaused means the simulated controller is paused(or will be paused).
+    ControllerPaused    StepPhase = "ControllerPaused"
+	// ControllerCompleted means the current running simulated controller no longer do anything with the current cluster state.
+    ControllerCompleted StepPhase = "ControllerCompleted"
+	// StepCompleted means the controller is preparing to move to the next step.
+	StepCompleted StepPhase = "Finished"
```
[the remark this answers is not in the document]
sanposhiho: Let me fix. Maybe the difference is from tab vs space * 4.

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,368 +64,539 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
-  // DeleteOperation indicates the operation to delete a resource.
-  // When use DeleteOperation, Operation should be "Delete".
-  DeleteOperation *DeleteOperation
-  // DoneOperation indicates the operation to mark the scenario as DONE.
-  // When use DoneOperation, Operation should be "Done".
-  DoneOperation   *DoneOperation
+type ControllerSet struct {
+	// Enabled specifies controllers that should be enabled.
+	// +listType=atomic
+	Enabled  []Controller `json:"enabled"`
```
everpeace, line 104: The order of Controllers affects the execution orders of SimulatedControllers?? If so, we should explain this explicitly.

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,368 +64,539 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
+We use `WaitForControllerFinish` to wait for all controllers to finish their work to deal with (1).
+
+##### 4. the simulated controller starts
+
+ScenarioStep: {X, 0}
+StepPhase: ControllerRunning
+
+The simulated controllers have been stopped until now.
+
+Here, start one of the simulated controllers.
```
everpeace, line 456: In which order if the Scenario has multiple simulated controllers?? We should describe it here.

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,368 +64,539 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
-The users want to see how their customized scheduler behaves in the worst case scenario.
+When the scenario controller is started and finds the Scenario which phase is "Running", the controller changes the status "Failed" with updating the `.status.message` like "the controller restarted while the Scenario was running".
 
-##### Solution
+#### Prohibitions and restrictions
 
-Even when a scenario is running, users can add events to that scenario. 
-So, in this case, they can add events that are most worst case for the scheduler by looking at the simulation results and the resources status at that point.
+The scenarios run one by one, and multiple scenarios are never run simultaneously. 
+That means the scenario controller will run the following Scenario after the current running Scenario becomes "Failed" or "Succeeded".
```
everpeace, line 595: If a user wanted to run hundreds of Scenarios, how to order them??  Is there any interface to specify the order for executing Scenarios??

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,368 +64,539 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
 
-## Alternatives
+In addition, the following actions are prohibited during scenario execution. The scenario result will be unstable or invalid if these actions are performed.
+- change the scheduler configuration via simulator API.
+- create/delete/edit any resources.
```
everpeace, line 599:
```suggestion
- create/delete/edit any resources manually.
```

everpeace (changes requested):
Wow, Thank you very much for the tough rewriting!! I left several comments for documenting.  LGTM on the concept and the ScenarioStep execution model.

keps/140-scenario-based-simulation/README.md
sanposhiho: > These are run one by one in the same order specified in Enabled field.

It's described in L95, but yeah, I'd add the note for it here as well.

keps/140-scenario-based-simulation/README.md
sanposhiho: Yea, it's described in L95.

everpeace: oh, I missed it. Thanks.

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,368 +64,539 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
 - the function to aggregate changes in allocation rate of the entire cluster.
 - the function to aggregate changes in resource utilization for each Node.
-- the function to aggregate data by Pod.
 - the generic iterator function that users can aggregate custom values.
 - (Do you have any other idea? Tell us!)
 
-By putting only the minimum simple information in ScenarioResult and providing functions to change it into a user-friendly structs, many data structures can be supported in the future without any changes to API.
-
-#### The case kube-apiserver have Scenarios when the controller start to run
+By putting only the minimum simple information in ScenarioResult and providing functions to change it into a user-friendly structs, we can support many data structures without any changes to Scenario API.
```
196Ikuchil, line 560:
```suggestion
By putting only the minimum simple information in ScenarioResult and providing functions to change it into a user-friendly struct, we can support many data structures without any changes to Scenario API.
```

196Ikuchil (commented):
Thank you! Overall., LGTM👍
Only some nits.

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,368 +64,539 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
-  NormalizedScore      int64
-  // FinalScore has score plugin’s final score calculated by normalizing with NormalizedScore and applied Score plugin weight. 
-  FinalScore           int64
+The Kubernetes controller has the desired state.
+For example, the ReplicaSet controller. It's working on managing the number of Pods to satisfy all replicaset's `.spec.replicas`.
+
+Some controllers in this world refer to something other than k8s resources like external metrics and real time to determine their ideal state. But, they cannot be used in our Scenario. In other words, **controllers that refer only to resources on k8s can be used in Scenario.**
+This is because the scenario controller cannot manage something other than k8s resources, and unexpected operations based on them may happen during the Scenario, which we want to avoid.
+
+The following controllers cannot be used in Scenario, for example:
```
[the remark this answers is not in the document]
sanposhiho: > This is because the scenario controller cannot manage something other than k8s resources, and unexpected operations based on them may happen during the Scenario, which we want to avoid.

Isn't the explanation above enough? What do you expect it?

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,368 +64,539 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
+- controllers that use external metrics.
+  - e.g., the cluster autoscaler that scales Nodes based on external metrics.
+- controllers that do something based on real time.
+  - e.g., the cronjob controller that creates Job based on real time.
+
+All controllers that refer only to k8s resources won't do anything if there is no change on k8s resources state.
+In other words, all controllers may do something only when someone performs a k8s resource operation.
+
+#### Expect when controllers finish working
```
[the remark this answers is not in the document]
sanposhiho: Renamed to `How the simulator knows when a cluster state gets converged by the controller?`.

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,368 +64,539 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
+> In Kubernetes, controllers are control loops that watch the state of your cluster, then make or request changes where needed. Each controller tries to move the current cluster state closer to the desired state.
+https://kubernetes.io/docs/concepts/architecture/controller/
+
+The simulated controller always checks ScenarioStep at the end of its loop. And if StepPhase is ControllerPaused, the controller stops working. 
+
+StepPhase was changed to ControllerStopped at (5), and the simulated controller should be stopped at the end of that loop.
+
+See also [# How to stop the simulated controllers loop](#How-to-stop-the-simulated-controllers-loop).
+
+##### 7. wait preparing controllers to handle resource changes
```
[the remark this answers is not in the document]
sanposhiho: It's correct. It wanna say that wait "preparing controllers" to handle resource changes.

keps/140-scenario-based-simulation/README.md
sanposhiho: > If a user wanted to run hundreds of Scenarios 

I'm thinking they are gonna run in the order in which they are created.

> Is there any interface to specify the order for executing Scenarios??

I'd say we don't need to consider it at least for now, but it may be useful to have.

sanposhiho:
@everpeace @196Ikuchil Sorry for toooooo late updating. I think I addressed/replied all your reviews. Please retake a look 🙏

196Ikuchil:
@sanposhiho 
Thanks! LGTM

@everpeace 
What do you think? I think it is time to start implementing this. 
And we would probably notice some new things as we begin it.

keps/140-scenario-based-simulation/README.md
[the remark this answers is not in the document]
everpeace: oh, ok. I think it's enough too. thanks.

keps/140-scenario-based-simulation/README.md
```diff
@@ -43,368 +64,539 @@ Therefore, this kep proposes to define the scenario **as CRD**. All clients, inc
...
 }
 ```
 
+For example, the ReplicaSet controller. `WaitForControllerFinish` detects wait for either of the following:
+  - Become ReplicaSet.spec.replicas == {the number of Pods matching ReplicaSet.spec.selector} (the ideal state for the ReplicaSet controller)
+  - Become Replicaset.status.conditions has ReplicaFailure true (cannot do any more to move the cluster state closer to the ideal state)
+
+Talk about why we need to expect it in the later section.
+
+#### Simulated controllers and preparing controllers
```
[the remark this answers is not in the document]
everpeace: ah, those look nice.

everpeace:
@sanposhiho LGTM! Thanks for your very hard work!

> What do you think? I think it is time to start implementing this.

@196Ikuchil Yes. This PR is now time to merge. Let's proceed. Thanks for your review either!!

196Ikuchil:
@sanposhiho 
Thanks for your hard work! 
/lgtm

PR #217, opened 2022-08-16
support external IP in docker-compose.yaml

amewayne:
#### What type of PR is this?

/kind documentation
/kind bug

This PR is aimed to fix an issue after installing `
kube-scheduler-simulator` through current docker-compose file. 

If I deploy the simulator on a remote server, it will return errors when requesting `simulator-server` due to wrong request URL, which is quite confusing.

image

image

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ You can use the following command.
 make docker_build_and_up
 ```
 
-Then, you can access the simulator with http://localhost:3000
+Then, you can access the simulator with http://localhost:3000. If you want to deploy the simulator on a remote server and access it via a specific IP (e.g: like http://10.0.0.1:3000/), please make sure that you have executed `export SIMULATOR_EXTERNAL_IP=your.server.ip` before running `docker-compose up -d`.
 
 Note: Insufficient memory allocation may cause problems in building the image.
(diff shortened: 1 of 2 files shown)


196Ikuchil:
Hi, @amewayne! 
Thank you for your PR.
I checked and run correctly in local.
LGTM

Properly speaking, it should be possible to specify a URL that includes the scheme and port, and to specify different external IPs for the simulator server and the frontend(because there is a possibility to deploy these on different machines).
But I think that this `docker-compose.yaml` is more required to be able to deploy quickly somewhere for all users, not to setting details.

WDYT, @sanposhiho?

/ok-to-test
/lgtm
/assign @sanposhiho

amewayne:
@196Ikuchil Exactly, it would be much better able to specify a URL in `docker-compose.yml` in the same way. But at this point, being able to to specifying an external IP is adequate to meet most of the users' need.

sanposhiho:
This PR is valid. 
I think those who wanna deploy server and web UI in different machines don't use docker-compose I guess.
Thanks @amewayne!

/remove-kind bug
/kind feature
/retitle support external IP in docker-compose.yaml
/approve

PR #232, opened 2022-09-24
Fix some error messages to avoid `TypeError` occurring.

196Ikuchil:
#### What type of PR is this?
/area web
/kind bug

Fix some error messages to avoid `TypeError` occurring.

These XXXKey variables are a `Symbol`type and can't covert to a `string` type.
It is unlikely that this throw line will be reached, but as it is now, the following error will occur.
```
TypeError: Cannot convert a Symbol value to a string
```

Fixes #

/ok-to-test

### DIFF
--- a/web/components/ResourceAddButton.vue
+++ b/web/components/ResourceAddButton.vue
@@ -58,32 +58,32 @@ export default defineComponent({
 
     const podstore = inject(PodStoreKey);
     if (!podstore) {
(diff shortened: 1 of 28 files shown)


yanghesong (approved):
/lgtm

sanposhiho (approved):
/lgtm
/approve

Thanks!

PR #239, opened 2022-10-03
Fix custom plugin example of nodenumber

### LINKED ISSUE #238: Custom plugin example NodeNumber is not working as expected
When I was testing the simulator with custom plugin using the example of how-to-use-custom-plugins, I found it not working as described.

I setup three nodes, and then try to new a pod, in the web, the new pod is in Unscheduled state:
image

when I checked the log of simulator, I found following:
image

I checked the code, it's the plugin failed to set the pre score state because the last char of the pod name is not a digit:
image

I was wandering is it better to ignore the error if the pod name is not ended with a digit?

/kind documentation

panli889:
#### What type of PR is this?

/area simulator
/kind bug

The custom plugin example may not run as expected, this PR will fix it.

Fixes #238


sanposhiho (approved):
Thanks!
/lgtm
/approve
/ok-to-test

PR #242, opened 2022-10-12
Fix a bug that finalscore in pod anno does not comply with weight in sche-config

wsq704252212:
#### What type of PR is this?
/area simulator
/kind bug

The final score doesn't change after editing the weight of score plugins in scheduler configuration, because result store doesn't read weight from scheduler configuration.

Moreover, following the guide in [https://github.com/kubernetes-sigs/kube-scheduler-simulator/tree/master/simulator/docs/how-to-use-custom-plugins](url), if user doesn't set weight in `OutOfTreeScorePlugins()`, the final score of custom plugin will be always 0 (because weight in scheduler configuration does not take effect).

Fixes #
I fixed this bug by the way. Should I open an issue first?

### Before:

before1
before2
before3

### After:

after1
after2
after3

### DIFF
--- a/simulator/scheduler/plugin/plugins.go
+++ b/simulator/scheduler/plugin/plugins.go
@@ -2,12 +2,14 @@ package plugin
 
 import (
 	"encoding/json"
+	"strings"
 
 	"golang.org/x/xerrors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	"k8s.io/kube-scheduler/config/v1beta2"
+	schedulerConfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	schedulerRuntime "k8s.io/kubernetes/pkg/scheduler/framework/runtime"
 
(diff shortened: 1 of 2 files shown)


sanposhiho:
@wsq704252212 
Thanks so much! will review it later.

/ok-to-test
/triage accepted
/priority important-soon

simulator/scheduler/plugin/plugins.go
```diff
@@ -271,3 +265,32 @@ func registeredFilterScorePlugins() ([]v1beta2.Plugin, error) {
...
+		if p.Weight != nil {
+			scorePluginWeight[p.Name] = *p.Weight
+		}
+	}
+
+	// TODO: support multi-scheduler
+	// overwrite scorePluginWeight by weight of enabled plugin in scheduler-config
+	if len(cfg.Profiles) == 0 {
+		return nil, xerrors.Errorf("get enabled score plugins: %w", err)
+	}
```
sanposhiho, line 287: I believe a given `cfg` should already have the all weight we should use.

Because ↓here merges the default weight and user-specified weight.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/3c90c5e2aeb7d9e22ed0e8b591b059dea5dced4b/simulator/scheduler/plugin/plugins.go#L174-L184

So, we can eliminate these lines, wdyt?

sanposhiho: I think you can eliminate other lines(L272 - L281) as well since `cfg` should have all plugins score.
The code I pointed in [the comment](https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/242#discussion_r1003581313) already merges the weights for all plugins (given by users + the default weight in case users don't set any weight for the plugin.)

wsq704252212: If I eliminate lines(L272 - L281) and user set default weights ↓here for their custom plugins，these weights will not take effect.

> https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/3c90c5e2aeb7d9e22ed0e8b591b059dea5dced4b/simulator/scheduler/config/plugin.go#L56-L60

But I think eliminating L272-L281 is reasonable. Because user should explicitly set weight in scheduler configuraton, not in the code. 

I will remove lines(L272 - L281)

sanposhiho:
@wsq704252212 Hello. Could you rebase it when have time?

wsq704252212:
ok, I will rebase it in next week

sanposhiho (approved):
Thanks @wsq704252212 

/approve
/lgtm

Issue #245, opened 2022-10-24
Create the component config


sanposhiho:
/kind feature
/kind cleanup
/area simulator

Currently, the simulator is configured via environment variables.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/simulator/docs/env-variables.md

I'd like to create the component config to configure the simulator and deprecate those envs in the future.

sanposhiho:
/remove-lifecycle stale
/lifecycle frozen

Still important. Will be implemented by #249

PR #249, opened 2022-11-14
Create the component config for simulator.(#245)

yanghesong:
Create the component config in simulator.

Signed-off-by: yanghesong <hesong.yang@foxmail.com>

#### What type of PR is this?
/kind feature
/kind cleanup
/area simulator

#### Which issue(s) this PR fixes:

Fixes #245

#### Special notes for your reviewer:
/cc @sanposhiho

PR #258, opened 2022-12-15
Create user config file for web

yanghesong:
#### What type of PR is this?
/area web

Create user config file for web. Refer to #245.

Fixes #

### DIFF
--- a/simulator/docs/env-variables.md
+++ b/simulator/docs/env-variables.md
@@ -38,6 +38,8 @@ still a beta feature.
 
 ## For Web UI
 
+This part is deprecated. Please refer to the config file [user.config.js](./../../web/user.config.js).
+
 `KUBE_API_SERVER_URL`: This is the kube-apiserver URL. Its default
 value is `http://localhost:3131`.
 
--- a/web/nuxt.config.js
+++ b/web/nuxt.config.js
@@ -1,4 +1,5 @@
 import fs from "fs";
+import userConfig from "./user.config.js";
 
 const getTemplateEnv = (hostEnv) => {
   return hostEnv !== "production"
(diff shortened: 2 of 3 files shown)


sanposhiho:
In https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/245, I was thinking about only the simulator (not web). Do you think introducing it benefits us?

yanghesong:
> In #245, I was thinking about only the simulator (not web). Do you think introducing it benefits us?
1. It is more convenient to start the web project manually, without having to manually set the environment variables.
2. Configuration files support more complex data structures, but the data structures supported by environment variables are relatively simple.
3. The configuration can be separated from the Dockerfile, and the user does not have to directly modify the Dockerfile (directly modifying the Dockerfile increases the possibility of problems), which is more user-friendly.

196Ikuchil:
Umm, I'm not sure it is not good. -1 😅
Just a thought...
If distributed as a docker container, it will be more difficult for the user to reflect the configuration.
It would also not be good to have more configuration files that need to be managed in addition to the Dockerfile.

sanposhiho:
@yanghesong @196Ikuchil 

I don't disagree to have the config file for web. We can pass it by mounting the file if we use the web in the container.
But, first of all, I believe the strength of having the configuration as a file is we can have the versioning on it. (like you're working on the simulator https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/249.)

With the versioning, we can indicate for users like "this config is now alpha and may get an unexpected breaking change" "this config is stable now and we don't give any breaking changes on it (as long as it's in the same major version)" etc.

So, to sum up:
- for me, creating the config file for web as well kinda makes sense.
- But, if doing so, we should give a versioning concept on it somehow.

What do you think?

196Ikuchil:
@sanposhiho @yanghesong 
Thank you for your comment!
This is my opinion that should not be implemented until this is actually needed. (Maybe I just don't fully understand the need for this feature...) Might it be unnecessary as a result? 

And then, if we need it, there are two issues to be discussed.
1. I think it is enough to use `.env`.
- `runtimeConfig` supports `.env` file. [link.](https://nuxtjs.ir/guide/runtime-config#-code-env-code-support) I think this fixing will be the same as mounting the `.env` and versioning. (but we also need to consider that the `.env` file should be on `.gitignore` in general.) This is mean we already have this solution, which is to add and manage `.env`.
2. Did this fix work?
- Sorry I haven't tested it locally yet, but I think this fix would compile including `user.config.js`? In that case, I feel that there is no need to bother with `runtimeConfig` and a different implementation would be fine.

Thanks!

Issue #109, opened 2022-02-14
Implement functions to fetch results from pod annotations


sanposhiho:
To use this simulator as a library, we needed to fetch scheduling results from pod annotations. 
In annotation, the results are stored as json, so the functions should unmarshal the json in the pod annotation and return the result as some useful type.

/kind feature

PR #259, opened 2022-12-20
API to fetch pod annotations and more

markstur:
Closes: 109

Adding endpoint /namespaces/:namespace/pods/:name/metadata/annotations/:annotation

This allows fetching of a single annotation from a specific pod in a specific namespace. The path (.../metadata/annotation...) is intended to mirror the apimachinery API to avoid re-inventing especially assuming this might be expanded.

Given the leveraging of apimachinery, it made sense to add some additional endpoints:

* /namespaces/:namespace/pods/:name/metadata/annotations - when no :annotation is added (or is blank) all annotations are returned
* /namespaces/:namespace/pods/:name - pod details are returned for namspace/pod
* /namespaces/:namespace/pods - pod details are returned for all pods in namespace

Unmarshalling of annotations:

Since the specific goal is to help access scheduler-simulator annotations in a usable format, we attempt to unmarshal specific annotations.  A specific scheduler-simulator/* annotation will become map[string]map[string]string.

In addition, when :annotation = "scheduler-simulator" or "scheduler-simulator/" we can conclude that all the simulator annotations for this pod should be returned and unmarshalled into map[string]map[string]map[string]string.

Other cases dump all the metadata and do not unmarshal annotations to avoid errors.

Signed-off-by: Mark Sturdevant <mark.sturdevant@ibm.com>

Fixes #

### DIFF
--- a/simulator/docs/api-samples/v1/pod_annotations.md
+++ b/simulator/docs/api-samples/v1/pod_annotations.md
@@ -1,0 +1,573 @@
+# Pod Annotations
+
+The simulator server provides access to unmarshalled scheduler-simulator annotations with simple GET requests producing JSON output.
+
+In addition, all pod information can be accessed as provided by apimachinery. The simulator server provides endpoints for a specific pod, all pods in a namespace, or all pods in all namespaces.
+
+## Get a specific scheduler-simulator annotation for a specific pod
+
+* For example: **"scheduler-simulator/filter-result"**
+* **Annotations are unmarshalled**
+* Data structure comes from apimachinery
+
+### curl command
+```bash
+curl http://localhost:1212/api/v1/namespaces/default/pods/pod-w64pz/metadata/annotations/scheduler-simulator/filter-result
+```
+
+### Output
+
+```json
+{
+ "node-pprb7": {
+  "AzureDiskLimits": "passed",
+  "EBSLimits": "passed",
+  "GCEPDLimits": "passed",
+  "InterPodAffinity": "passed",
+  "NodeAffinity": "passed",
+  "NodeName": "passed",
+  "NodePorts": "passed",
+  "NodeResourcesFit": "passed",
+  "NodeUnschedulable": "passed",
+  "NodeVolumeLimits": "passed",
+  "PodTopologySpread": "passed",
+  "TaintToleration": "passed",
+  "VolumeBinding": "passed",
+  "VolumeRestrictions": "passed",
+  "VolumeZone": "passed"
+ }
+}
+```
+
+## Get all scheduler-simulator annotations for a specific pod
+
+* **Annotations are unmarshalled**
+* Data structure comes from apimachinery
+* This only includes annotations that have a "scheduler-simulator/" prefix.
+
+### curl command
+```bash
+curl http://localhost:1212/api/v1/namespaces/default/pods/pod-w64pz/metadata/annotations/scheduler-simulator
+```
+
+### Output
+
+```json
+{
+ "scheduler-simulator/filter-result": {
+  "node-pprb7": {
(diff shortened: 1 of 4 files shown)


196Ikuchil:
@markstur 
Hello! Thank you for your PR!

Our simulator allows us to communicate with api-server on port `3131`, which means that we could already get and list pods. So I think the necessary endpoint is only below. (Cc. @sanposhiho )

> /namespaces/:namespace/pods/:name/metadata/annotations/:annotation
/namespaces/:namespace/pods/:name/metadata/annotations

(It is possible that `/namespaces/:namespace/pods/:name/metadata` will also be needed someday.)



And I would like to discuss [this implementation](https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/259/files#diff-006c17a771b65662e4879448848d81d726b1aee689d12b645339f9a31d73be5cR71).
If you put a string of conditions such as `scheduler-simulator` directly here, you may forget to modify it if the annotation name is changed. (this is also a problem on our simulator that the current simulator does not yet define such strings in a one place.)
Also, the string `scheduler-simulator` should be used as a prefix in the future, so it would be better to handle it in `c.Param()`. WDYT?

sanposhiho:
> Closes: 109

@markstur Do you mean it's for https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/109?
In #109, I mean we want to have just functions (not HTTP APIs) like:

```
func GetFilteringResult(pod *v1.Pod, pluginName, nodeName string) string {
}

func GetScoreingResult(pod *v1.Pod, pluginName, nodeName string) string {
}
```

markstur:
> @markstur Hello! Thank you for your PR!
> 
> Our simulator allows us to communicate with api-server on port `3131`, which means that we could already get and list pods. So I think the necessary endpoint is only below. (Cc. @sanposhiho )
> 
> > /namespaces/:namespace/pods/:name/metadata/annotations/:annotation
> > /namespaces/:namespace/pods/:name/metadata/annotations
> 
> (It is possible that `/namespaces/:namespace/pods/:name/metadata` will also be needed someday.)
> 
> And I would like to discuss [this implementation](https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/259/files#diff-006c17a771b65662e4879448848d81d726b1aee689d12b645339f9a31d73be5cR71). If you put a string of conditions such as `scheduler-simulator` directly here, you may forget to modify it if the annotation name is changed. (this is also a problem on our simulator that the current simulator does not yet define such strings in a one place.) Also, the string `scheduler-simulator` should be used as a prefix in the future, so it would be better to handle it in `c.Param()`. WDYT?

Thanks @196Ikuchil ...

1. Yes I added some "pass-thru" convenience, that is probably unnecessary creep. It seemed useful for completeness, but I can remove those.

2. The hard-coded prefix "scheduler-simulator" problem... So there are 2 things going on here.
    * Since we put a / in our annotation names, using our prefix is handy to get all our annotations. But hard-coding is bad, so I'd probably change it to a query like: /namespaces/:namespace/pods/:name/metadata/annotations?prefix=scheduler-simulator. That would allow the scheduler-simulator filtering to work and retrieve all matches without a hard-coded prefix.
    * But I only wanted to do the unmarshal to our specific structure for our recognized annotations.  Well, I already added a default which tries and falls-back to the encoded string.  So I could just use that.

My understanding was that this issue wanted the unmarshal pretty stuff, but I can't guarantee all the annotations will conform.

So I can make those changes.

markstur:
@sanposhiho So you want clients to use the simulator module, config some api-server that is active, and call those functions?  I guess in simulator/simulator.go?

Should the web API be abandoned?

markstur:
Seems this is not the right solution.  I'll close the old PR.

PR #260, opened 2023-01-04
Configure kube-apiserver to allow privileged containers.

enisoc:
This is equivalent to passing `--allow-privileged` to kube-apiserver. It is needed for any simulations that include privileged Pods, and it shouldn't have any effect on simulations without such Pods.

Fixes #


sanposhiho (commented):
Thanks.

/lgtm
/approve

PR #262, opened 2023-01-09
fix: grant execute permission to etcd.sh

Gekko0114:
#### What type of PR is this?
grant execute permission to etcd.sh.

/area simulator
/kind bug
/kind deprecation

I thought it is weird that etcd.sh doesn't have execute permission.
On the other hand, I know that developers don't run etcd.sh directly.
So if this PR is not needed, please abandon this PR.

Thank you very much for this amazing repository!
I am impressed and would like to contribute it.
This PR is just a small change, but I would like to contribute it more in the future.


196Ikuchil:
Hello! @Gekko0114 
Thank you for your PR.
Sorry, but I think that this change is not necessary, currently.
As you know, this script will not be called directly or executed via make command. Giving unnecessary authority could lead to risk.

We look forward to your further contribution! 😄 

@sanposhiho 
Could I close this PR?

Gekko0114:
I see, thanks for your comment!

sanposhiho:
Yep, seems we don't need it.

/close

PR #272, opened 2023-03-03
Fix crash due to concurrent Pod mutation

enisoc:
We should make a copy instead of directly mutating the Pod that was sent by the informer.

This fixes a crash in the scheduler (https://github.com/kubernetes/kubernetes/issues/116226), which was trying to concurrently read the same Pod object we mutated.


sanposhiho (approved):
Thanks very much!

/lgtm
/approve

PR #317, opened 2023-08-06
Fix to considering the order of import of PriorityClass on Pods

### LINKED ISSUE #303: Can't import Pod correctly from yaml on WebUI
Follow the below steps and get these error logs.
1. Create Pod
2. Create PriorityClass
3. Export
4. Reset
5. Import
6. Failed ↓
```sh
E0531 01:12:28.949622   12592 snapshot.go:58] failed to load all resources: failed to apply():
    sigs.k8s.io/kube-scheduler-simulator/simulator/snapshot.(*Service).Load
        /Users/i196/go_project/kube-scheduler-simulator/simulator/snapshot/snapshot.go:261
  - apply resources:
    sigs.k8s.io/kube-scheduler-simulator/simulator/snapshot.(*Service).apply
        /Users/i196/go_project/kube-scheduler-simulator/simulator/snapshot/snapshot.go:229
  - apply Pod:
    sigs.k8s.io/kube-scheduler-simulator/simulator/snapshot.(*Service).applyPods.func1
        /Users/i196/go_project/kube-scheduler-simulator/simulator/snapshot/snapshot.go:545
  - apply pod:
    sigs.k8s.io/kube-scheduler-simulator/simulator/pod.(*Service).Apply
        /Users/i196/go_project/kube-scheduler-simulator/simulator/pod/pod.go:51
  - pods "pod-hnjhk" is forbidden: the integer value of priority (0) must not be provided in pod spec; priority admission controller computed 1000 from the given PriorityClass name
```

I think this is due to the fact that PriorityClass is Imported(Apply) first.

/kind bug

196Ikuchil:
#### What type of PR is this?
/area simulator
/kind bug

The Snapshot feature was not able to take PriorityClass into account when importing pods. This meant that it was not working properly.

- Fixed to import pods before importing PriorityClass
- Fixed creating dummy PriorityClass when importing pods. (It will be removed after completion)

Fixes #303 

### DIFF
--- a/simulator/go.mod
+++ b/simulator/go.mod
@@ -105,6 +105,7 @@ require (
 	github.com/google/gofuzz v1.1.0 // indirect
 	github.com/google/uuid v1.1.2 // indirect
 	github.com/googleapis/gax-go/v2 v2.1.1 // indirect
+	github.com/gotidy/ptr v1.4.0 // indirect
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.7.0 // indirect
 	github.com/imdario/mergo v0.3.6 // indirect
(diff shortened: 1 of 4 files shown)


sanposhiho:
Instead of creating dammy PC, can we make it simpler by like this?

- split applyPods into applyUnprioritizedPodsI() and applyPrioritizedPods()
- applyUnprioritizedPods() → applyPcs() → errgrp.Wait() → applyPrioritizedPods()

196Ikuchil:
@sanposhiho 
Thanks! Splitting is a good idea for me ✋ 

> applyUnprioritizedPods() → applyPcs() → errgrp.Wait() → applyPrioritizedPods()

But, this idea↑ doesn't solve this problem↓.

> It is possible that a PriorityClass with the same name but different value existed previously. (maybe edge case)
https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/303#issuecomment-1666452855

In this case, all pods with the same priorityClass are overwritten with the newer priority value.

Issue #251, opened 2022-11-22
Integrate kwok into simulator


sanposhiho:
In the simulator, we just run almost pure kube-apiserver. The simulator doesn't have much controllers, and something seems actually weird to users sometimes. (like some fields in Pods and Nodes don't get changed, etc)
And, for example, https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/243 directly affects the scheduling results, which we must avoid as a simulator. 

I feel we can almost resolve them by integrating [Kwok](https://github.com/kubernetes-sigs/kwok) which watches Node and Pods and behaves like kubelet. We can expect Pods and Nodes to behave more like a real cluster but without running any real instances or containers.

/kind feature
/triage accepted
/area simulator
/priority important-longterm

wzshiming:
Cool, I think the cluster created by `kwokctl` should not start the scheduler and `kube-scheduler-simulator` should take care of the scheduling, so that kwok should be able to be integrated. What do you think?

sanposhiho:
> kwokctl should not start the scheduler and kube-scheduler-simulator should take care of the scheduling

Yes. In other words, I feel the simulator should depend on all other parts like running controller-manager, kube-apiserver, etc. 
Once kwokctl gets the ability to disable scheduler, we can refactor the simulator's scheduler to run on top of kwok, and this repository can focus on the simulating related stuff.

sivchari:
I'm interesting at this issue. If you would like, I would like discuss the integration, too.

wzshiming:
I'm back, sorry for the wait 🙏.
[The Kwok v0.1.0](https://github.com/kubernetes-sigs/kwok/releases/tag/v0.1.0) has been released, which supports the ability to disable the default scheduler, and an all-in-one image that I particularly like.

## Use all-in-one image create cluster

``` bash
docker run --rm -it --env KWOK_DISABLE_KUBE_SCHEDULER=true  -p 8080:8080 registry.k8s.io/kwok/cluster:v1.26.0
```

## Use kwokctl create cluster

``` bash
kwokctl create cluster --disable-kube-scheduler=true
```

sanposhiho:
Thanks @wzshiming for sharing. 
One thing to confirm, does Kwok support the CRD? I mean, can we use CR as a usual kube-apiserver when we apply CRD on Kwok?

wzshiming:
Of course, it is used just like a normal cluster.

[...]

PR #322, opened 2023-10-21
Skip create api server

sivchari:
#### What type of PR is this?

/assign @sivchari 

/area simulator

kube-scheduler-simulator absolutely create a Kubernetes API Server even if KubeAPIServerURL of configuration value is not empty. So this changes make kube-scheduler-simulator skip to create new Kubernetes API Server if KubeAPIServerURL is not empty.

Related with #251 

### DIFF
--- a/simulator/simulator.go
+++ b/simulator/simulator.go
@@ -9,7 +9,9 @@ import (
 
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"golang.org/x/xerrors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	clientset "k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/config"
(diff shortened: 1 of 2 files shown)


simulator/simulator.go
```diff
@@ -35,11 +37,23 @@ func startSimulator() error {
 		return xerrors.Errorf("get config: %w", err)
 	}
 
-	restclientCfg, apiShutdown, err := k8sapiserver.StartAPIServer(cfg.KubeAPIServerURL, cfg.EtcdURL, cfg.CorsAllowedOriginList)
-	if err != nil {
-		return xerrors.Errorf("start API server: %w", err)
+	kubeAPIServerURL := cfg.KubeAPIServerURL
```
sanposhiho, line 40: KubeAPIServerURL has the default value. 
So, you gotta change something if you'd like to reuse this.

https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/0be3194647ec86ec8e0b3befb94c015ddf366c0c/simulator/config/config.go#L139-L156

Or, probably just adding `EnableKubeAPIServer` parameter to the config is the simplest way?

sanposhiho (commented):
/ok-to-test

simulator/simulator.go
sivchari: > Or, probably just adding EnableKubeAPIServer parameter to the config is the simplest way?

It looks good to me. How about like this ? We remove KubeAPIServer URL default value or remove KubeAPIServerURL field , add a new receiver method to return the string of KubeAPIServerURL. I think it's simple way. What do you think ?

sanposhiho: I don't quite get what you mean. You mean creating another env named `KubeAPIServerURL` and remove default values from `KUBE_API_PORT` and `KUBE_API_HOST`? (If wrong, please elaborate.

sivchari: Sorry, I don't know what I was getting at, too. Please forget it once.
I think it's good way to add new env named ExternalKubeAPIServerURL. If is's set, a simulator uses external kube-api-server. If not, a simulator uses internal kube-api-server which creates itself. What do you think ?

sanposhiho: sg

sanposhiho:
/hold

We have to discuss fully the idea of integrating kwok without installing CLI in the issue. ([ref](https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/251#issuecomment-1837124903))
If we can, this PR wouldn't be needed anymore in the first place.

sanposhiho:
/close

It's no longer valid and should be replaced with the PR to create kube-apiserver via [kubernetes-sigs/e2e-framework/support/kwok](https://github.com/kubernetes-sigs/e2e-framework/blob/c61338a3773c2b64550d0e46037bfe541fa54645/support/kwok/kwok.go#L135).

PR #333, opened 2024-02-06
migrate the internal kube-apiserver to kwok

### LINKED ISSUE #251: Integrate kwok into simulator

sivchari:
#### What type of PR is this?

/area simulator

/kind feature
/kind cleanup

Before this PR, we should manage not only kube-scheduler, but also kube-apiserver and etcd.
We wanna focus on kube-scheduler and reduce the time when we maintain various components. 

Therefore, this aims to integrate kwok into simulator that simulates Nodes and Clusters..
If this PR is merged, all we have to do  only maintain scheduler-simulator. 

Fixes #251 


docker-compose-local.yml
```diff
@@ -5,39 +5,30 @@ services:
...
-    image: quay.io/coreos/etcd:v3.4.26
-    container_name: simulator-etcd
-    restart: always
-    volumes:
-      - simulator-etcd-data:/var/lib/etcd
-    command: etcd --advertise-client-urls http://simulator-etcd:2379 --data-dir /var/lib/etcd --listen-client-urls http://0.0.0.0:2379 --initial-cluster-state new --initial-cluster-token tkn
-    networks:
-      - simulator-internal-network
-volumes:
-  simulator-etcd-data:
```
wzshiming, line 40: I think it's better to create clusters directly from existing mirrors, so that the simulator doesn't need to be concerned with the creation and deletion of clusters at all, and only needs to be externally supplied with etcd ports.


``` yaml
  simulator-cluster:
    image: registry.k8s.io/kwok/cluster:v0.5.0-k8s.v1.29.0
    container_name: simulator-cluster
    restart: always
    ports:
      - "8080:8080"
    environment:
      - KWOK_KUBE_APISERVER_PORT=8080
    volumes:
      - simulator-etcd-data:/var/lib/etcd
      - ./kwokctl.yaml:/root/.kwok/kwok.yaml
```

```yaml
# kwokctl.yaml
kind: KwokctlConfiguration
apiVersion: config.kwok.x-k8s.io/v1alpha1
options:
  etcdPort: 2379
  etcdPrefix: /kube-scheduler-simulator
  disableKubeScheduler: true
componentsPatches:
- name: kube-apiserver
  extraArgs:
  - key: cors-allowed-origins
    value: ^*$
```

sivchari: How can we know if cluster is ready ?

sivchari: https://t.co/VGv4GwKzmm

Sorry for the clutter. Although I tried to fix these codes, it couldn't catch the watchResources response. it's suspended and I don't have no idea why it's caused. Do you have some idea ?

sivchari:
@sanposhiho 
I'm ready for review. 

@wzshiming 
Thanks for your so kind support.

sanposhiho:
/retitle migrate the internal kube-apiserver to kwok

@sivchari Thanks, I'll take a look this weekend.

simulator/config.yaml
```diff
@@ -24,13 +24,9 @@ corsAllowedOriginList:
...
-# This is the host of kube-apiserver which the simulator
-# starts internally. Its default value is 127.0.0.1.
-kubeApiHost: "127.0.0.1"
-
-# This is the port of kube-apiserver. Its default value is 3131.
-kubeApiPort: 3131
-
+# This is the url of kube-apiserver.
+# This variable is used to connect to the user cluster's kube-apiserver.
+kubeApiServerUrl: ""
```
sanposhiho, line 29: There should be the document about the config. Please update it as well.

simulator/simulator.go
```diff
@@ -35,19 +34,10 @@ func startSimulator() error {
...
-
-	client := clientset.NewForConfigOrDie(restclientCfg)
-
-	ctrlerShutdown, err := controller.RunController(client, restclientCfg)
-	if err != nil {
-		return xerrors.Errorf("start controllers: %w", err)
+	restCfg := &rest.Config{
+		Host: cfg.KubeAPIServerURL,
 	}
-	defer ctrlerShutdown()
```
sanposhiho, line 50: Veeery happy to see they're removed finally!

simulator/simulator.go
```diff
@@ -65,10 +55,7 @@ func startSimulator() error {
 		return xerrors.Errorf("create an etcd client: %w", err)
 	}
 
-	// need to sleep here to make all controllers create initial resources. (like "system-" priorityclass.)
-	time.Sleep(1 * time.Second)
```
sanposhiho, line 69: How can we know whether the kwok cluster is ready or not.

simulator/reset/reset.go
```diff
@@ -41,7 +40,7 @@ func NewResetService(
 		schedService: schedService,
 	}
 
-	result, err := etcdClient.Get(context.Background(), "/"+k8sapiserver.EtcdPrefix, clientv3.WithPrefix())
+	result, err := etcdClient.Get(context.Background(), "/kube-scheduler-simulator", clientv3.WithPrefix())
```
sanposhiho, line 43: make `"/kube-scheduler-simulator"` `const` var.

simulator/config/config.go
```diff
@@ -136,23 +139,15 @@ func getPort() (int, error) {
...
-		if p == "" {
-			p = "3131"
-		}
-	}
-
-	h := os.Getenv("KUBE_API_HOST")
-	if h == "" {
-		h = configYaml.KubeAPIHost
-		if h == "" {
-			h = "127.0.0.1"
```
sanposhiho, line 152: Ah, we no longer need to keep these configurations actually. But, we cannot immediately remove them because it'd be a breaking change.

For now, please add a note on the document of config that they're old configuration and will be removed very soon.

sanposhiho (commented):
Very rough review for now

simulator/simulator.go
sivchari: I'd try to get the `kube-system namespace` to check whether cluster is ready.
I think it's better way as Namespace resource exist from the begining.
What do you think ?

wzshiming (commented):
A few comments and no hard opinions.

docker-compose-local.yml
```diff
@@ -5,13 +5,10 @@ services:
     container_name: simulator-server
     environment:
       - PORT=1212
-      - KUBE_SCHEDULER_SIMULATOR_ETCD_URL=http://simulator-etcd:2379
-      - CORS_ALLOWED_ORIGIN_LIST=http://${SIMULATOR_EXTERNAL_IP:-localhost}:3000
-      - KUBE_API_HOST=0.0.0.0
-      - KUBE_API_PORT=3131
+      - KUBE_SCHEDULER_SIMULATOR_ETCD_URL=http://simulator-cluster:2379
```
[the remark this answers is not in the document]
sivchari: It sounds simple and clear for me to define as ETCD_URL

kwok.yaml
```diff
@@ -0,0 +1,12 @@
+kind: KwokctlConfiguration
+apiVersion: config.kwok.x-k8s.io/v1alpha1
+options:
+  kubeApiserverPort: 8080
```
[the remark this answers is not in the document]
sivchari: It's so important to keep backward compatibility, so I'd change the port from 8080 to 3131.

simulator/simulator.go
```diff
@@ -65,10 +56,11 @@ func startSimulator() error {
...
+	if _, err := client.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{}); err != nil {
+		return xerrors.Errorf("get kube-system namespace: %w", err)
+	}
```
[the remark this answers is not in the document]
sivchari: How about using [Forever](https://pkg.go.dev/k8s.io/apimachinery@v0.29.1/pkg/util/wait#Forever) ?

simulator/util/consts.go
```diff
@@ -0,0 +1,3 @@
+package util
+
+const EtcdPrefix = "/kube-scheduler-simulator"
```
sanposhiho, line 3: It's weird to put in util... Can we create another package named `const` to store all this kind of static values?

simulator/simulator.go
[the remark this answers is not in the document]
sanposhiho: Let's go with [`UntilWithContext`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#UntilWithContext) so that it can be canceled. We can set the timeout with `context.WithDeadline`.

simulator/docs/environment-variables.md
```diff
@@ -28,6 +32,9 @@ origin for `CORS_ALLOWED_ORIGIN_LIST`.
 resources". This variable is used to find Kubeconfig required to
 access your cluster for importing resources to scheduler simulator.
 
+`KUBE_APISEVER_URL`: This is the URL of kube-apiserver which the
+simulator starts internally. Its default value is `http://{simulator-cluster-ip}:8080`.
```
sanposhiho, line 36:
```suggestion
simulator uses internally. Its default value is `http://{simulator-cluster-ip}:8080`.
```

sanposhiho, line 36: > Its default value is `http://{simulator-cluster-ip}:8080`.

I don't see where we set this. Anyways, I guess we should make it required configuration, shouldn't we. Because we no longer start kube-apiserver internally and the simulator cannot run without kube-apiserver outside

simulator/docs/environment-variables.md
```diff
@@ -5,6 +5,10 @@ Until deprecation, the simulator will read the configuration in the environment
 if the environment variable is not set, it will read the configuration in the configuration file.
 For config file, please refer to the simulator [config.yaml](./../config.yaml).
 
+> [!WARNING]
+> `KUBE_API_HOST` and `KUBE_API_PORT` are deprecated. They will be removed in the near future.
+> Please use `KUBE_APISEVER_URL` instead.
```
sanposhiho, line 10:
```suggestion
> To reduce the maintenance burden, the simulator no longer run up kube-apiserver internally 
> and we made the simulator require the kube-apiserver outside.
> `KUBE_API_HOST` and `KUBE_API_PORT` are deprecated. They will be removed in the near future, meanwhile they're just ignored.
```

simulator/docs/environment-variables.md
```diff
@@ -28,6 +32,9 @@ origin for `CORS_ALLOWED_ORIGIN_LIST`.
 resources". This variable is used to find Kubeconfig required to
 access your cluster for importing resources to scheduler simulator.
 
+`KUBE_APISEVER_URL`: This is the URL of kube-apiserver which the
+simulator starts internally. Its default value is `http://{simulator-cluster-ip}:8080`.
+
 `KUBE_API_HOST`: This is the host of kube-apiserver which the
```
sanposhiho, line 38: Modify the explanation of `KUBE_API_HOST` and `KUBE_API_PORT` as well.

simulator/docs/environment-variables.md
sanposhiho, line 1: Regarding the doc, you also have to modify:
- https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/simulator/docs/kube-apiserver.md
- https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/simulator/docs/how-it-works.md

Also please modify the top level README as well so that we clearly mention the need of kube-apiserver outside and recommend people use Kwok outside the simulator.

simulator/simulator.go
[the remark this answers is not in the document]
sivchari: Thanks. @sanposhiho

simulator/docs/environment-variables.md
sivchari: OK, I agree with your opinion. I''ll make this value required value.

simulator/docs/environment-variables.md
sivchari: > meanwhile they're just ignored.

I have a simple question. Do you plan to ignore `KUBE_API_HOST` and `KUBE_API_PORT` until both values is removed ?

https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/333/files#diff-1cb29cb30fedf9a3629f7ebbaca8a15317fa8fa4a03b2aa07b81ba3438d42759L139-R144

The current changes prioritize `KUBE_APISERVER_URL` if the values is set. In other hand, if the values isn't set, we use `KUBE_API_HOST` and `KUBE_API_PORT` to keep backward compatibility. So quoted sentence may confuse to someone that use kube-scheduler-simulator.  Or do I should change the implementation not to use both configuration values ? What do you think ?

sanposhiho: > if the values isn't set, we use KUBE_API_HOST and KUBE_API_PORT to keep backward compatibility.

It changes the meaning of these configurations, which is not great. 
The previous meaning was they're the values to configure the kube-apiserver which the simulator internally starts and we don't have such internal kube-apiserver anymore. So, I'd say just ignoring them is the simplest idea here.

sivchari: OK, I understand what you said. 
It's exactly right, so I'll modify this explanation.

sivchari: Thanks @sanposhiho

docker-compose-local.yml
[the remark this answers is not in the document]
sanposhiho: We don't have to worry about the env. The configuration thru env is deprecated and we'll remove them eventually. So, the current way of configuring is basically thru scheduler config file, which has EtcdURL instead of this messy long title.

simulator/util/consts.go
sivchari: I agree with that it's weird. But if all kind of static values moves to `consts` package, I guess it becomes difficult for maintainers to read. Because, I think the each static value has already been located properly. But I don't have better idea to store this value to suitable place. Or this value is only used in reset package, so should we write in it ?

simulator/docs/how-it-works.md
sanposhiho, line 1: Similarly, Please modify [README](https://github.com/kubernetes-sigs/kube-scheduler-simulator/README.md) too.

simulator/simulator.go
```diff
@@ -65,10 +57,20 @@ func startSimulator() error {
 		return xerrors.Errorf("create an etcd client: %w", err)
 	}
 
-	// need to sleep here to make all controllers create initial resources. (like "system-" priorityclass.)
-	time.Sleep(1 * time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	defer cancel()
+	wait.UntilWithContext(ctx, func(ctx context.Context) {
```
sanposhiho, line 62: Sorry, I misled in a previous comment.
We should use PollUntilContextTimeout rather, otherwise `wait.UntilWithContext` wouldn't panic or error out if reaching timeout seemingly.
https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollUntilContextTimeout

README.md
```diff
@@ -41,12 +41,12 @@ We have several components:
...
-Simulator is kube-apiserver with some controllers + debuggable scheduler + the HTTP server which mainly for the web UI.
+Simulator is composed of debuggable scheduler + the HTTP server which mainly for the web UI.
 
 There are several ways to integrate your scheduler into the simulator.
 See [integrate-your-scheduler.md](simulator/docs/integrate-your-scheduler.md).
 
-You can create any resources by any ways (kubectl, k8s client library, or web UI described next).
+You can create any resources by communication with kube-apiserver outside via any clients (e.g. kubectl, k8s client library, or web UI described next).
 And when you create Pods, 
 Pods will be scheduled by the [debuggable scheduler](./simulator/docs/debuggable-scheduler.md),
```
sanposhiho, line 51:
```suggestion
Simulator runs with a fake cluster powered by [KWOK](https://github.com/kubernetes-sigs/kwok)

You can create any resources in KWOK cluster via any clients (e.g. kubectl, k8s client library, or web UI described next).
And when you create Pods, 
Pods will be scheduled by the [debuggable scheduler](./simulator/docs/debuggable-scheduler.md),
```

simulator/simulator.go
```diff
@@ -65,10 +57,22 @@ func startSimulator() error {
...
 
-	// need to sleep here to make all controllers create initial resources. (like "system-" priorityclass.)
-	time.Sleep(1 * time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+	err = wait.PollUntilContextCancel(ctx, time.Second*5, true, func(ctx context.Context) (bool, error) {
+		_, err := client.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{})
+		if err != nil {
+			klog.Infof("waiting for kube-system namespace to be ready: %v", err)
+			return false, err
```
sanposhiho, line 66: I guess if we here return err, it'd immediately fail without retry. Let's just return nill as we log error in L65.

```suggestion
			return false, nil
```

simulator/config/config.go
```diff
@@ -137,22 +138,26 @@ func getPort() (int, error) {
 
 // getKubeAPIServerURL gets KubeAPIServerURL from environment variable first, if empty from the config file.
 func getKubeAPIServerURL() string {
+	url := os.Getenv("KUBE_APISERVER_URL")
+	if url == "" && configYaml.KubeAPIServerURL != "" {
+		return configYaml.KubeAPIServerURL
+	}
 	p := os.Getenv("KUBE_API_PORT")
```
sanposhiho, line 145: Let's remove L145-160, but without removing KubeAPIPort and KubeAPIHost from `simulator/config/v1alpha1/types.go`. We can just ignore deprecated envs (as I said before, removing them would be a breaking change that we should avoid though

README.md
```diff
@@ -41,12 +41,12 @@ We have several components:
```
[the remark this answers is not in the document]
sivchari: I see. I'd revise it.

sivchari:
I've updated PR, thanks.

docker-compose-local.yml
```diff
@@ -27,17 +24,19 @@ services:
...
+  simulator-cluster:
+    image: registry.k8s.io/kwok/cluster:v0.5.0-k8s.v1.29.0
+    container_name: simulator-cluster
     restart: always
+    ports:
+      - "3131:3131"
     volumes:
       - simulator-etcd-data:/var/lib/etcd
-    command: etcd --advertise-client-urls http://simulator-etcd:2379 --data-dir /var/lib/etcd --listen-client-urls http://0.0.0.0:2379 --initial-cluster-state new --initial-cluster-token tkn
+      - ./kwok.yaml:/root/.kwok/kwok.yaml
```
wzshiming, line 35:
```suggestion
      - ./kwok.yaml:/root/.kwok/kwok.yaml
      environment:
      - KWOK_KUBE_APISERVER_PORT=3131
```

kwok.yaml
```diff
@@ -0,0 +1,12 @@
+kind: KwokctlConfiguration
+apiVersion: config.kwok.x-k8s.io/v1alpha1
+options:
+  kubeApiserverPort: 3131
```
wzshiming, line 4:
```suggestion
```

wzshiming (commented):
Currents, the image [startup script is hardcoded](https://github.com/kubernetes-sigs/kwok/blob/7c7daed6f7bf866c6726f0fb8871a7a96f9ca6a8/images/cluster/entrypoint.sh#L51-L57), the only way to set this is through the environment variable.
and the next release (kubernetes-sigs/kwok#999) will support `kubeApiserverInsecurePort` to instead of this

sivchari:
@wzshiming 
Thanks!! I've checked the script of kwok's entry point and the plan of next release.
Once, I decided to use a env to connect the cluster in KWOK.

docker-compose-local.yml
```diff
@@ -5,13 +5,10 @@ services:
     container_name: simulator-server
     environment:
       - PORT=1212
-      - KUBE_SCHEDULER_SIMULATOR_ETCD_URL=http://simulator-etcd:2379
-      - CORS_ALLOWED_ORIGIN_LIST=http://${SIMULATOR_EXTERNAL_IP:-localhost}:3000
```
sanposhiho, line 9: It's still needed as it's being used by simulator's server (not kube-apiserver) too.

docker-compose.yml
```diff
@@ -5,13 +5,10 @@ services:
     container_name: simulator-server
     environment:
       - PORT=1212
-      - KUBE_SCHEDULER_SIMULATOR_ETCD_URL=http://simulator-etcd:2379
-      - CORS_ALLOWED_ORIGIN_LIST=http://${SIMULATOR_EXTERNAL_IP:-localhost}:3000
```
sanposhiho, line 9: ditto

kwok.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: config.kwok.x-k8s.io/v1alpha1
+options:
+  etcdPort: 2379
+  etcdPrefix: /kube-scheduler-simulator
+  disableKubeScheduler: true
+componentsPatches:
+- name: kube-apiserver
+  extraArgs:
+  - key: cors-allowed-origins
+    value: ^*$
```
sanposhiho, line 11: Seemingly it's not working. @wzshiming Any idea?

You should be able to reproduce it by `make docker_up_local` and visit localhost:3000 (web ui).

スクリーンショット 2024-03-22 23 47 24

wzshiming: I did some debugging

in v0.1.1
```
curl http://localhost:3131/api/v1/namespaces//pods

{
    "kind": "PodList",
    "apiVersion": "v1",
    "metadata": {
        "resourceVersion": "215"
    },
    "items": []
}
```

in this pr
```
curl http://localhost:3131/api/v1/namespaces//pods

# 301 to /api/v1/namespaces/pods

{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "namespaces \"pods\" not found",
  "reason": "NotFound",
  "details": {
    "name": "pods",
    "kind": "namespaces"
  },
  "code": 404
}
```

Getting all pods might look like this

```
curl http://localhost:3131/api/v1/pods
```

sanposhiho: Interesting but that one may be a separated another issue from CORS one that I shew? 
I think CORS issue is specific to web browser.

wzshiming: No, get fails, but it can create it. It's not a CORS problem.

I tried booting all the way up to 0.1.1 first, bypassing the initialization, and then later uninstalling 0.1.1, and then booting this PR version, and it worked.

wzshiming: Or let's try putting node or ns first, and we'll know if it's a CORS problem.

wzshiming: image

Strangely, after I rebuild the frontend, the path works, but the simulator's CORS is broken.

wzshiming: Oh, I typed the address. It should be http://localhost:3000/. It's working now. just rebuild frontend.

sanposhiho: Well, so my CORS issue isn't reproducible. I have to recheck things on my side again then.

sivchari: スクリーンショット 2024-03-26 9 00 22

Thank you for investigating this matter. I tried to reproduce on my environment, but I couldn't.

docker-compose-local.yml
sanposhiho: nvm, we don't need because simulator/config.yaml has it.

simulator/docs/environment-variables.md
sanposhiho, line 1: We forget about the doc in:
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/f0faa06ec7773723073ad04c4bb3383dbbd8d3ff/simulator/docs/simulator-server-config.md

Can you
- remove the mention to kubeApiHost and kubeApiPort
- mention kubeApiServerUrl

over there?

sanposhiho (commented):
Two doc changes, otherwise LGTM

sivchari:
@sanposhiho 
I've revised it, thanks.

sanposhiho (approved):
/lgtm
/approve

@sivchari This is epic, thanks for your all effort!

sanposhiho:
Also thanks @wzshiming for your all support!

PR #335, opened 2024-02-18
PoC: implement syncer

### LINKED ISSUE #327: mimicube: the tool to synchronically replicate cluster state from a real cluster to a fake one for the scheduler testing
/assign
/kind feature
/assign

---

This issue proposes the feature to keep syncing the state between a prod cluster and a fake cluster in the simulator.

### background

Testing the scheduler is a complex challenge. There are countless patterns of operations executed within a cluster, making it impractical to anticipate every scenario with a finite number of tests. More often than not, bugs are discovered only when the scheduler is deployed in an actual cluster.

Having a development or sandbox environment for testing the scheduler—or, indeed, any Kubernetes controllers—is a common practice. However, this approach falls short of capturing all the potential scenarios that might arise in a production cluster. It’s an inevitable truth that a development cluster never sees the exact same use or exhibits the same behavior as its production counterpart, with notable differences in workload sizes and scaling dynamics.

### User story

We have a custom scheduler which has a co-scheduling feature. 
We want to test it in a cluster that gets similar resources as our production cluster. But, our production cluster is much bigger than our development cluster and it's unrealistic to catch all bugs there.

#### Resources to sync

We shouldn't simply do that, we have to think about what to sync and what not to.

All resources involved in the scheduling should be synced. 
And, we should make it configurable to select which resources to sync, given everyone could have a different scheduler plugin which schedules Pods based on anything.

By default, we should sync:
- Pods
- Nodes
- PVs
- PVCs
- SC

### Scheduled Pods

We cannot simply sync all changes to Pods, because the real cluster has the scheduler, and it schedules all Pods in the cluster.
If we simply synced all changes to Pods, the scheduling result would also be synced. (and may conflicted with the decision of another scheduler which is in a fake cluster.)

So, we don't sync any of updated events to scheduled Pods. 
Pods are synced like:
1. In a real cluster, Pod-a is created
2. In a fake cluster, Pod-a is created. (synced)
3. In a real cluster, the scheduler schedules Pod-a to Node-a. We don't copy this change to a fake cluster.
4. In a fake cluster, the scheduler, which is different one from (3), schedules Pod-a to Node-x.

sanposhiho:
#### What type of PR is this?

/area simulator
/kind feature

Implement syncer which keep copying the resources from the external kube-apiserver to the simulator's one.
It's based on https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/333 so that it won't be hard to rebase in the future.

Fixes #327


utam0k:
We are happy to take over this PR, but could you tell me what its status is?

sanposhiho:
Feel free to just take over by cherry pick or whatever

PR #346, opened 2024-03-31
Use docker compose instead of docker-compose

utam0k:
#### What type of PR is this?

/area simulator
/kind cleanup

Nowadays, we don't need to install `docker-compose`. We can use `docker compose` instead of that.
Also, it means we use docker compose v2, however as far as I know, it doesn't affect our use cases.
https://docs.docker.com/compose/migrate/

None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,19 +41,19 @@ docker_build_front:
 
 .PHONY: docker_up
 docker_up:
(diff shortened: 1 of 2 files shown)


sanposhiho (approved):
/lgtm
/approve

PR #351, opened 2024-05-02
: handle the case when the config option is not specified

YamasouA:
#### What type of PR is this?
/area simulator
/kind bug

Resolve error when config flag is empty.

Fixes #


sanposhiho (approved):
Thanks!

/ok-to-test
/lgtm
/approve
/retitle: handle the case when the config option is not specified

PR #356, opened 2024-05-23
Support arm64

yeahdongcn:
#### What type of PR is this?
/area web
/area simulator
/area scenario
/kind feature

I want to run the simulator on Apple-Silicon Mac through `make docker_up` but the image only has `amd64` platform.
After checking the `Dockerfile` and `Makefile`, it turns out `amd64` is hard coded in `GOARCH`.
This PR introduces `USE_BUILDX` to control whether to enable `buildx` to build `arm64` local images. By default, `make docker_build` acts the same as before. `USE_BUILDX=1 make docker_build` will setup buildx parameters and build `arm64` images with buildx then export the images to docker.

Fixes #

Testing done (on my M1 MBP):
* USE_BUILDX=1 make docker_build -> ok
* make docker_up_local -> ok
* Access the simulator in http://localhost:3000 -> ok


sanposhiho (commented):
/ok-to-test

Will take a look hopefully within the weekend.

Makefile
```diff
@@ -1,3 +1,11 @@
+BUILD      ?= build
+USE_BUILDX ?=
+
+# Setup buildx flags
+ifneq ("$(USE_BUILDX)","")
+BUILD = buildx build --platform=linux/arm64 -o type=docker
```
sanposhiho, line 6: Do we need to fix it with `--platform=linux/arm64`? I suppose if we provide no `--platform`, the default value is decided based on your machine and `TARGETARCH` is set to arm64.

```suggestion
BUILD = buildx build -o type=docker
```

yeahdongcn: This is for `amd64` guys.

sanposhiho: Got it, so what you want to do is building the image on amd64 and run it arm64.

Makefile
```diff
@@ -1,3 +1,11 @@
...
+endif
```
sanposhiho, line 7:
```suggestion
BUILD      ?= build

USE_BUILDX ?=
BUILDX_PLATFORM ?= linux/amd64

# Setup buildx flags
ifneq ("$(USE_BUILDX)","")
BUILD = buildx build --platform=$(BUILDX_PLATFORM) -o type=docker
endif
```

yeahdongcn: This looks unnecessary.

`USE_BUILDX` is introduced for building `arm64` images on both `amd64` and `arm64` machines (the behavior is the same). Furthermore, you can push the image to some registry or create manifests for supporting multi-archs.

In your new change, `USE_BUILDX=1 make docker_build` still builds `amd64` image.

sanposhiho: So, you have to use `USE_BUILDX=1 BUILDX_PLATFORM =linux/arm64 make docker_build` for your usecase.

My point is to make `USE_BUILDX` straightforward. `USE_BUILDX` literally only means whether using buildx for docker build or not. From its name, people cannot relate this flag to the platform. 
Our current default (w/o `USE_BUILDX`) is linux/amd64 and hence we shouldn't break it with a flag named like this.

So.. then should we rename `USE_BUILDX` to something like `BUILD_FOR_ARM64`? No.
From its functionality perspective too, `USE_BUILDX` (or `BUILD_FOR_ARM64` if we renamed) is only useful for your specific usecase, but not for other people who want to build for a platform apart from arm64.

yeahdongcn: A new commit is added to use `uname -m` to get the host arch. Please take a look. Thanks.

sanposhiho (approved):
/lgtm
/approve

Thanks

Issue #363, opened 2024-08-20
Replace golang/mock with uber-go/mock


saza-ku:
gomock is deprecated. Its maintained fork is uber-go/mock.

ref: https://github.com/golang/mock

> This repo and tool are no longer maintained. Please see [go.uber.org/mock](https://github.com/uber/mock) for a maintained fork instead.

chansuke:
I'd like to work on this issue.

PR #365, opened 2024-08-21
Rename `ImportClusterResourceService` to `OneshotClusterResourceImporter`

saza-ku:
#### What type of PR is this?

/area simulator
/kind cleanup

If we add the syncing service (#327), it has two features to import resources of a real cluster, one-shot importing and continuous syncing.
So the current naming of `ImportClusterResouceService` would be confusing.

This PR renames `ImportClusterResourceService` to `OneshotClusterResourceImporter`. This will avoid the confusion.

Part of #327 

We should also rename `mock_clusterresourceimporter`, but this is generated by gomock so we need to replace golang/mock with uber-go/mock (#363). This PR once leaves it.


utam0k (commented):
/lgtm from my eyes

The code itself looks good. But I'd like to know @sanposhiho's opinion.

ordovicia (approved):

sanposhiho (approved):
/lgtm
/approve

PR #367, opened 2024-08-22
Implement syncer

### LINKED ISSUE #327: mimicube: the tool to synchronically replicate cluster state from a real cluster to a fake one for the scheduler testing

saza-ku:
#### What type of PR is this?
/area simulator
/kind feature

This PR implements syncer that continuously sync the simulator cluster's resources to an external cluster's ones.

Fixes #327

This PR is too large, but it's difficult to break down it because of testing. Please see this commit by commit: 

- implement syncer: https://github.com/kubernetes-sigs/kube-scheduler-simulator/commit/82a49436fff4845c0812244f0e9f244d0ff9cecb
- add syncer to the di container and invoke it from the entry point: https://github.com/kubernetes-sigs/kube-scheduler-simulator/commit/c844916117a2203526ecba417d5f5c92e69353e8

### DIFF
--- a/simulator/cmd/simulator/simulator.go
+++ b/simulator/cmd/simulator/simulator.go
@@ -11,13 +11,24 @@ import (
 	"golang.org/x/xerrors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/dynamic"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/config"
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/server"
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/server/di"
+	"sigs.k8s.io/kube-scheduler-simulator/simulator/syncer"
+)
+
+const (
+	kubeAPIServerPollInterval = 5 * time.Second
+	kubeAPIServerReadyTimeout = 2 * time.Minute
+	importTimeout             = 2 * time.Minute
 )
 
 // entry point.
@@ -40,12 +51,22 @@ func startSimulator() error {
 		Host: cfg.KubeAPIServerURL,
 	}
 	client := clientset.NewForConfigOrDie(restCfg)
+	dynamicClient := dynamic.NewForConfigOrDie(restCfg)
+	discoverClient := discovery.NewDiscoveryClient(client.RESTClient())
+	cachedDiscoveryClient := memory.NewMemCacheClient(discoverClient)
+	restMapper := restmapper.NewDeferredDiscoveryRESTMapper(cachedDiscoveryClient)
 
 	importClusterResourceClient := &clientset.Clientset{}
-	if cfg.ExternalImportEnabled {
+	var importClusterDynamicClient dynamic.Interface
+	if cfg.ExternalImportEnabled || cfg.ResourceSyncEnabled {
 		importClusterResourceClient, err = clientset.NewForConfig(cfg.ExternalKubeClientCfg)
 		if err != nil {
 			return xerrors.Errorf("creates a new Clientset for the ExternalKubeClientCfg: %w", err)
+		}
+
+		importClusterDynamicClient, err = dynamic.NewForConfig(cfg.ExternalKubeClientCfg)
+		if err != nil {
+			return xerrors.Errorf("creates a new dynamic Clientset for the ExternalKubeClientCfg: %w", err)
 		}
 	}
 
@@ -57,9 +78,10 @@ func startSimulator() error {
 		return xerrors.Errorf("create an etcd client: %w", err)
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	err = wait.PollUntilContextCancel(ctx, time.Second*5, true, func(ctx context.Context) (bool, error) {
+
+	err = wait.PollUntilContextTimeout(ctx, kubeAPIServerPollInterval, kubeAPIServerReadyTimeout, true, func(ctx context.Context) (bool, error) {
 		_, err := client.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{})
 		if err != nil {
 			klog.Infof("waiting for kube-system namespace to be ready: %v", err)
@@ -72,7 +94,7 @@ func startSimulator() error {
 		return xerrors.Errorf("kubeapi-server is not ready: %w", err)
 	}
 
-	dic, err := di.NewDIContainer(client, etcdclient, restCfg, cfg.InitialSchedulerCfg, cfg.ExternalImportEnabled, importClusterResourceClient, cfg.ExternalSchedulerEnabled, cfg.Port)
+	dic, err := di.NewDIContainer(client, dynamicClient, restMapper, etcdclient, restCfg, cfg.InitialSchedulerCfg, cfg.ExternalImportEnabled, cfg.ResourceSyncEnabled, importClusterResourceClient, importClusterDynamicClient, cfg.ExternalSchedulerEnabled, cfg.Port, syncer.Options{})
 	if err != nil {
 		return xerrors.Errorf("create di container: %w", err)
 	}
@@ -86,10 +108,18 @@ func startSimulator() error {
 	// If ExternalImportEnabled is enabled, the simulator import resources
 	// from the target cluster that indicated by the `KUBECONFIG`.
 	if cfg.ExternalImportEnabled {
-		ctx := context.Background()
 		// This must be called after `StartScheduler`
-		if err := dic.OneshotClusterResourceImporter().ImportClusterResources(ctx); err != nil {
+		timeoutCtx, timeoutCancel := context.WithTimeout(ctx, importTimeout)
+		defer timeoutCancel()
+		if err := dic.OneshotClusterResourceImporter().ImportClusterResources(timeoutCtx); err != nil {
 			return xerrors.Errorf("import from the target cluster: %w", err)
+		}
+	}
+
+	if cfg.ResourceSyncEnabled {
+		// Start the resource syncer to sync resources from the target cluster.
+		if err = dic.ResourceSyncer().Run(ctx); err != nil {
+			return xerrors.Errorf("start syncing: %w", err)
 		}
 	}
 
--- a/simulator/config.yaml
+++ b/simulator/config.yaml
@@ -19,7 +19,8 @@ etcdURL: "http://127.0.0.1:2379"
 corsAllowedOriginList:
   - "http://localhost:3000"
 
-# This is for the beta feature "Importing cluster's resources".
+# This is for the beta feature "One-shot importing cluster's resources"
+# and "Continuous syncing cluster's resources".
 # This variable is used to find Kubeconfig required to access your
 # cluster for importing resources to scheduler simulator.
 kubeConfig: "/kubeconfig.yaml"
@@ -35,9 +36,17 @@ kubeApiServerUrl: ""
 kubeSchedulerConfigPath: ""
 
 # This variable indicates whether the simulator will
-# import resources from an user cluster's or not.
-# Note, this is still a beta feature.
+# import resources from a user cluster specified by kubeConfig.
+# Note that it only imports the resources once when the simulator is started.
+# You cannot make both externalImportEnabled and resourceSyncEnabled true because those features would be conflicted.
+# This is still a beta feature.
 externalImportEnabled: false
+
+# This variable indicates whether the simulator will
+# keep syncing resources from an user cluster's or not.
+# You cannot make both externalImportEnabled and resourceSyncEnabled true because those features would be conflicted.
+# Note, this is still a beta feature.
+resourceSyncEnabled: false
 
 # This variable indicates whether an external scheduler
 # is used.
--- a/simulator/config/config.go
+++ b/simulator/config/config.go
@@ -31,10 +31,13 @@ type Config struct {
 	KubeAPIServerURL      string
 	EtcdURL               string
 	CorsAllowedOriginList []string
-	// ExternalImportEnabled indicates whether the simulator will import resources from an target cluster or not.
+	// ExternalImportEnabled indicates whether the simulator will import resources from a target cluster once
+	// when it's started.
 	ExternalImportEnabled bool
+	// ExternalImportEnabled indicates whether the simulator will keep syncing resources from a target cluster.
+	ResourceSyncEnabled bool
 	// ExternalKubeClientCfg is KubeConfig to get resources from external cluster.
-	// This field is non-empty only when ExternalImportEnabled == true.
+	// This field should be set when ExternalImportEnabled == true or ResourceSyncEnabled == true.
 	ExternalKubeClientCfg *rest.Config
 	InitialSchedulerCfg   *configv1.KubeSchedulerConfiguration
 	// ExternalSchedulerEnabled indicates whether an external scheduler is enabled.
@@ -48,6 +51,8 @@ const (
 )
 
 // NewConfig gets some settings from config file or environment variables.
+//
+//nolint:cyclop
 func NewConfig() (*Config, error) {
 	if err := LoadYamlConfig(defaultFilePath); err != nil {
 		return nil, err
@@ -74,8 +79,12 @@ func NewConfig() (*Config, error) {
 	}
 
 	externalimportenabled := getExternalImportEnabled()
+	resourceSyncEnabled := getResourceSyncEnabled()
 	externalKubeClientCfg := &rest.Config{}
-	if externalimportenabled {
+	if externalimportenabled && resourceSyncEnabled {
+		return nil, xerrors.Errorf("externalImportEnabled and resourceSyncEnabled cannot be used simultaneously.")
+	}
+	if externalimportenabled || resourceSyncEnabled {
 		externalKubeClientCfg, err = clientcmd.BuildConfigFromFlags("", configYaml.KubeConfig)
 		if err != nil {
 			return nil, xerrors.Errorf("get kube clientconfig: %w", err)
@@ -98,6 +107,7 @@ func NewConfig() (*Config, error) {
 		ExternalImportEnabled:    externalimportenabled,
 		ExternalKubeClientCfg:    externalKubeClientCfg,
 		ExternalSchedulerEnabled: externalSchedEnabled,
+		ResourceSyncEnabled:      resourceSyncEnabled,
 	}, nil
 }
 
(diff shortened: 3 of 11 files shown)


utam0k:
Since these commits were based on @sanposhiho's PR, please add him as `Co-Author` at the first commit in this PR.
https://docs.github.com/ja/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

saza-ku:
> Since these commits were based on @sanposhiho's PR, please add him as `Co-Author` at the first commit in this PR.

Added!

simulator/cmd/simulator/simulator.go
```diff
@@ -40,13 +44,26 @@ func startSimulator() error {
 		Host: cfg.KubeAPIServerURL,
 	}
 	client := clientset.NewForConfigOrDie(restCfg)
+	dynamicClient, err := dynamic.NewForConfig(restCfg)
+	if err != nil {
+		return xerrors.Errorf("creates dynamic clientset: %w", err)
+	}
```
ordovicia, line 50: nit: panicking may be sufficient

```suggestion
	dynamicClient := dynamic.NewForConfigOrDie(restCfg)
```

simulator/cmd/simulator/simulator.go
```diff
@@ -57,9 +74,11 @@ func startSimulator() error {
...
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
-	defer cancel()
-	err = wait.PollUntilContextCancel(ctx, time.Second*5, true, func(ctx context.Context) (bool, error) {
+	ctx, cancel1 := context.WithCancel(context.Background())
+	defer cancel1()
+	timeoutctx, cancel2 := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel2()
+	err = wait.PollUntilContextCancel(timeoutctx, time.Second*5, true, func(ctx context.Context) (bool, error) {
```
ordovicia, line 81: `wait.PollUntilContextTimeout()` can be used here instead.

simulator/config/config.go
```diff
@@ -31,10 +31,13 @@ type Config struct {
 	KubeAPIServerURL      string
 	EtcdURL               string
 	CorsAllowedOriginList []string
-	// ExternalImportEnabled indicates whether the simulator will import resources from an target cluster or not.
+	// ExternalImportEnabled indicates whether the simulator will import resources from an target cluster once
```
ordovicia, line 34:
```suggestion
	// ExternalImportEnabled indicates whether the simulator will import resources from a target cluster once
```

simulator/config/config.go
```diff
@@ -31,10 +31,13 @@ type Config struct {
...
+	// when it's started.
 	ExternalImportEnabled bool
+	// ExternalImportEnabled indicates whether the simulator will keep syncing resources from an target cluster.
```
ordovicia, line 37:
```suggestion
	// ExternalImportEnabled indicates whether the simulator will keep syncing resources from a target cluster.
```

simulator/docs/import-cluster-resources.md
```diff
@@ -1,16 +1,31 @@
 ### [Beta] Import your real cluster's resources
 
-The simulator can import resources from your cluster.
+There are two ways to import resources from your cluster.
```
ordovicia, line 3: It would be better to mention that these two methods can be used simultaneously

simulator/syncer/syncer.go
```diff
@@ -0,0 +1,339 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+// GVRs is a list of GroupVersionResource that we want to sync.
+// Note that this order matters - When first importing resources, we want to sync namespaces first, then priorityclasses, storageclasses...
+var GVRs = []schema.GroupVersionResource{
```
ordovicia, line 21: This variable is public, so users can modify it.
In what cases would users need to change this variable?

simulator/syncer/syncer.go
```diff
@@ -0,0 +1,339 @@
...
+const (
+	Add Event = iota
+	Update
+)
+
+// MutatingFunctions is a map of GroupVersionResource to mutatingFunction.
+// This is public so that outside users can add their own mutating functions.
+var MutatingFunctions = map[schema.GroupVersionResource]MutatingFunction{
+	{Group: "", Version: "v1", Resource: "persistentvolumes"}: mutatePV,
+}
```
ordovicia, line 44: Modifying global variables are generally dangerous.
How about

1. Defining `defaultMutatingFunctions()` that returns the default list of mutating functions, and
2. Adding a parameter to `New()` so that users can pass a list

simulator/syncer/syncer.go
saza-ku: I think users would have a scheduler that schedules pods based on resources other than default ones.

> All resources involved in the scheduling should be synced.
> And, we should make it configurable to select which resources to sync, given everyone could have a different scheduler plugin which schedules Pods based on anything.

https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/327

ordovicia: Understood. Thank you for the explanation!

simulator/syncer/syncer.go
```diff
@@ -0,0 +1,336 @@
...
+
+func mutatePV(ctx context.Context, resource *unstructured.Unstructured, clients *Clients, _ Event) (*unstructured.Unstructured, error) {
+	var pv v1.PersistentVolume
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(resource.UnstructuredContent(), &pv)
+	if err != nil {
+		return nil, err
+	}
+
+	if pv.Status.Phase == "Bound" {
```
utam0k, line 239:
```suggestion
	if pv.Status.Phase == v1.VolumeBound {
```

simulator/syncer/syncer.go
```diff
@@ -0,0 +1,336 @@
...
+	)
+	if err != nil {
+		return xerrors.Errorf("failed to create resource: %v", err)
+	}
+
+	return nil
+}
+
+// removeMetadata removes the metadata from the resource.
+func removeMetadata(resource *unstructured.Unstructured) *unstructured.Unstructured {
```
utam0k, line 202: The method name `removeMetadata` might be misleading, as readers could interpret it as removing all metadata of the passed resource. Perhaps consider a diffirent name?
```suggestion
func removeUnnecessaryMetadata(resource *unstructured.Unstructured) *unstructured.Unstructured {
```

simulator/syncer/syncer.go
```diff
@@ -0,0 +1,336 @@
...
+var MutatingFunctions = map[schema.GroupVersionResource]MutatingFunction{
+	{Group: "", Version: "v1", Resource: "persistentvolumes"}: mutatePV,
+}
+
+// ValidatingFunctions is a map of GroupVersionResource to validatingFunction.
+// This is public so that outside users can add their own validating functions.
+var ValidatingFunctions = map[schema.GroupVersionResource]ValidatingFunction{
+	{Group: "", Version: "v1", Resource: "pods"}: validatePods,
+}
+
```
utam0k, line 51: Wouldn't it be easier to read the entire syncer logic if the logic regarding validate and muteting was written out to a separate file? WDYT?

simulator/syncer/syncer_test.go
```diff
@@ -0,0 +1,481 @@
...
+				if err != nil {
+					t.Fatalf("failed to convert pod to unstructured: %v", err)
+				}
+				unstructedPod := &unstructured.Unstructured{Object: p}
+				_, err = src.Resource(v1.Resource("pods").WithVersion("v1")).Namespace(pod.Namespace).Update(ctx, unstructedPod, metav1.UpdateOptions{})
+				if err != nil {
+					t.Fatalf("failed to update pod: %v", err)
+				}
+			}
+			time.Sleep(500 * time.Millisecond)
```
utam0k, line 436: I'd like to avoid `time.Sleep` because the time needed depends on the environment in which this test runs. Do you have any ideas to avoid it?

simulator/syncer/syncer.go
saza-ku: Definitely. The file is large.

simulator/syncer/syncer_test.go
saza-ku: `time.Sleep` between creating and updating, and one between updating and deleting might be unnecessary.

The problem is how to wait for the syncer to apply source cluster's resources to the destination cluster. This could be done by trying some times using `PollUntilContextTimeout`.

simulator/cmd/simulator/simulator.go
```diff
@@ -11,15 +11,25 @@ import (
...
 
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/config"
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/server"
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/server/di"
 )
 
+const (
+	kubeAPIServerPollInterval = 5 * time.Second
+	kubeAPIServerReadyTimeout = 2 * time.Minute
+	importTimeout             = 2 * time.Minute
```
ordovicia, line 30: It would be better to make this duration configurable via `config.yaml` because the required duration varies for the source environment.

simulator/docs/import-cluster-resources.md
```diff
@@ -1,16 +1,34 @@
...
+
+```yaml
+externalImportEnabled: true
+kubeConfig: "/path/to/your-cluster-kubeconfig"
+```
+
+#### Keep importing resources
+
+To use this, you need to follow these two steps in the scheduler configuration:
+- Set `true` to `externalImportSynced`.
```
ordovicia, line 22:
```suggestion
- Set `true` to `resourceSyncEnabled`. 
```

simulator/docs/import-cluster-resources.md
```diff
@@ -1,16 +1,34 @@
...
+#### Keep importing resources
+
+To use this, you need to follow these two steps in the scheduler configuration:
+- Set `true` to `externalImportSynced`. 
+- Set the path of the kubeconfig file for the your cluster to `KubeConfig`. 
+  - This feature only requires the read permission for resources.
 
 ```yaml
-externalImportEnabled: false
+externalSchedulerSynced: true
```
ordovicia, line 27:
```suggestion
resourceSyncEnabled: true
```

simulator/docs/import-cluster-resources.md
```diff
@@ -1,16 +1,34 @@
...
 kubeConfig: "/path/to/your-cluster-kubeconfig"
 ```
 
-Then, the simulator imports resources from your cluster once when it's initialized.
+> [!NOTE]
+> When you enable `externalImportSynced`, deleting resources of the simulator cluster would cause a problem of syncing. Please avoid it except when you are debugging the simulator. When you delete a resource for debugging, make sure you reboot the simulator and the fake source cluster.
```
ordovicia, line 32:
```suggestion
> When you enable `resourceSyncEnabled`, deleting resources of the simulator cluster would cause a problem of syncing. Please avoid it except when you are debugging the simulator. When you delete a resource for debugging, make sure you reboot the simulator and the fake source cluster.
```

simulator/cmd/simulator/simulator.go
saza-ku: Definitely. But I think another PR should fix it because the difference of this PR is already large. I added the issue: https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/369

ordovicia (commented):
LGTM. Nice work!
I want to wait for @sanposhiho and @utam0k reviews.

simulator/syncer/syncer.go
```diff
@@ -0,0 +1,280 @@
...
+	unstructObj, ok := newObj.(*unstructured.Unstructured)
+	if !ok {
+		klog.Error("Failed to convert runtime.Object to *unstructured.Unstructured")
+		return
+	}
+
+	err := s.updateResourceOnDestinationCluster(ctx, unstructObj)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			klog.Info("Skipped to update resource on destination: ", err)
```
sanposhiho, line 257:
```suggestion
			// We just ignore the not found error because the scheduler may preempt the Pods, or users may remove the resources for debugging.
			klog.Info("Skipped to update resource because we cannot find it in the destination cluster", "resource", klog.KObj(newObj))
```

simulator/syncer/syncer.go
```diff
@@ -0,0 +1,280 @@
...
+	unstructObj, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		klog.Error("Failed to convert runtime.Object to *unstructured.Unstructured")
+		return
+	}
+
+	err := s.deleteResourceOnDestinationCluster(ctx, unstructObj)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			klog.Info("Skipped to delete resource on destination: ", err)
```
sanposhiho, line 275: ditto

simulator/syncer/resource.go
```diff
@@ -0,0 +1,122 @@
...
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(resource.UnstructuredContent(), &pod)
+	if err != nil {
+		return nil, err
+	}
+
+	// Pods must have the default ServiceAccount because ServiceAccount is not synced.
+	// But static pods must not have a ServiceAccount.
+	if pod.Spec.ServiceAccountName != "" {
+		pod.Spec.ServiceAccountName = "default"
+	}
```
sanposhiho, line 93:
```suggestion
	
	// We always make it empty so that the default service account is used.
	// (We don't sync a service account because it's not related to the scheduling)
	pod.Spec.ServiceAccountName = ""
```

simulator/syncer/resource.go
```diff
@@ -0,0 +1,122 @@
...
+	pod.OwnerReferences = nil
+
+	modifiedUnstructed, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&pod)
+	return &unstructured.Unstructured{Object: modifiedUnstructed}, err
+}
+
+// filterPods checks if a pod is already scheduled when it's updated.
+// We only want to update pods that are not yet scheduled.
+func filterPods(_ context.Context, resource *unstructured.Unstructured, _ *Clients, event Event) (bool, error) {
+	if event != Update {
```
sanposhiho, line 105: It's more intuitive.

```suggestion
	if event == Add {
		// We always add a Pod, regardless it's scheduled or not.
```

simulator/syncer/resource.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	var pod v1.Pod
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(resource.UnstructuredContent(), &pod)
+	if err != nil {
+		return false, err
+	}
+
+	if pod.Spec.NodeName != "" {
+		klog.InfoS("Pod is scheduled. We ignore Pods already scheduled", "pod", pod.Name, "namespace", pod.Namespace)
```
sanposhiho, line 116:
```suggestion
		klog.InfoS("Pod is scheduled. We ignore updates to scheduled Pods", "pod", klog.KObj(pod))
```

simulator/config.yaml
```diff
@@ -39,6 +40,10 @@ kubeSchedulerConfigPath: ""
 # Note, this is still a beta feature.
 externalImportEnabled: false
 
+# This variable indicates whether the simulator will
+# keep syncing resources from an user cluster's or not.
+resourceSyncEnabled: false
```
sanposhiho, line 45:
```suggestion

# This variable indicates whether the simulator will
# import resources from a user cluster specified by kubeConfig.
# Note that it only imports the resources once when the simulator is started.
# You cannot make both externalImportEnabled and resourceSyncEnabled true because those features would be conflicted.
# This is still a beta feature.
externalImportEnabled: false

# This variable indicates whether the simulator will
# keep syncing resources from an user cluster's or not.
# You cannot make both externalImportEnabled and resourceSyncEnabled true because those features would be conflicted.
# Note, this is still a beta feature.
resourceSyncEnabled: false
```

simulator/docs/import-cluster-resources.md
```diff
@@ -1,16 +1,34 @@
...
 
 ```yaml
-externalImportEnabled: false
+resourceSyncEnabled: true
 kubeConfig: "/path/to/your-cluster-kubeconfig"
 ```
 
-Then, the simulator imports resources from your cluster once when it's initialized.
+> [!NOTE]
+> When you enable `resourceSyncEnabled`, deleting resources of the simulator cluster would cause a problem of syncing. Please avoid it except when you are debugging the simulator. When you delete a resource for debugging, make sure you reboot the simulator and the fake source cluster.
```
sanposhiho, line 32:
```suggestion
> When you enable `resourceSyncEnabled`, adding/updating/deleting resources directly in the simulator cluster could cause a problem of syncing. 
> You can do them for debugging etc purposes though, make sure you reboot the simulator and the fake source cluster afterward.
```

simulator/docs/simulator-server-config.md
sanposhiho, line 1: Add the same description change as simulator/config.yaml.

simulator/syncer/resource.go
saza-ku: If not static pods are applied without `ServiceAccountName`, the resource creation fails.

sanposhiho: Does it?

```bash
sanposhiho@Kenseis-MacBook-Pro workspace % cat pod.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
  - name: nginx
    image: nginx
    ports:
    - containerPort: 80
sanposhiho@Kenseis-MacBook-Pro workspace % kubectl apply -f pod.yaml
pod/nginx created
```

saza-ku: Just applying it via kubectl is okay. But applying not static pods without ServiceAccoutName in the syncer fails.

I don't know why it causes, but you can repro by the following steps:

- Do the fix you suggest.
- Start the simulator with the syncer.
- Create `foo` ServiceAccount and a pod that has `foo` ServiceAccount in the source simulator.

```
kwokctl kubectl apply -f - <<EOF
apiVersion: v1
kind: ServiceAccount
metadata:
  name: foo
---
apiVersion: v1
kind: Pod
metadata:
  name: nginx-1
spec:
  serviceAccountName: foo
  containers:
  - name: nginx
    image: nginx:1.14.2
EOF
```
- In the simulator cluster, you won't see the pod. The simulator server's log says:

```
E0828 09:24:53.405157       1 syncer.go:230] "Failed to create resource on destination cluster" err="failed to create resource: pods \"nginx-1\" is forbidden: error looking up service account default/foo: serviceaccount \"foo\" not found"
```

sanposhiho: What if we remove `ManagedFields` too?

saza-ku: > What if we remove `ManagedFields` too?

The result is the same.

saza-ku: This works:

```
	pod.Spec.ServiceAccountName = ""
	pod.Spec.DeprecatedServiceAccount = ""
```

saza-ku:
@sanposhiho I fixed it!: https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/367/#discussion_r1735449327

Issue #375, opened 2024-09-12
Issue in loading the kube scheduler configuration when not specifying it explicitly


aariv1:
Run into this issue recently after starting to use the release-simulator-v0.3.0. I'm creating this ticket to check if this is just something in my end or if there is an actual issue.

By default if I clone the kube-scheduler-simulator and run make docker_build_and_up, I will get an error like this:

```go
I0912 15:59:17.587909       1 scheduler.go:19] failed to build the debuggablescheduler command: failed to NewConfigs():
    sigs.k8s.io/kube-scheduler-simulator/simulator/pkg/debuggablescheduler.NewSchedulerCommand
        /go/src/simulator/pkg/debuggablescheduler/command.go:20
  - load scheduler config:
    sigs.k8s.io/kube-scheduler-simulator/simulator/pkg/debuggablescheduler.NewConfigs
        /go/src/simulator/pkg/debuggablescheduler/debuggable_scheduler.go:57
  - load scheduler config:
    sigs.k8s.io/kube-scheduler-simulator/simulator/pkg/debuggablescheduler.loadKubeSchedulerConfig
        /go/src/simulator/pkg/debuggablescheduler/debuggable_scheduler.go:149
  - open : no such file or directory
``` 

I was trying to understand where the default kube scheduler config is set and noticed that in debuggable_scheduler.go NewConfigs function we do this

```go
	configFile := flag.String("config", "", "")
	master := flag.String("master", "", "")
	// port indicates port number of the proxy server for Extenders.
	// This flag is debuggable_scheduler's own.
	port := flag.Int("proxyPort", 1212, "")
	flag.Parse()

	versionedcfg, err := loadKubeSchedulerConfig(configFile)
``` 
And in function loadKubeSchedulerConfig this

```go
func loadKubeSchedulerConfig(configFile *string) (*v1.KubeSchedulerConfiguration, error) {
	var versionedcfg *v1.KubeSchedulerConfiguration
	var err error
	if configFile == nil {
		versionedcfg, err = simulatorschedulerconfig.DefaultSchedulerConfig()
		if err != nil {
			return nil, xerrors.Errorf("get default scheduler config: %w", err)
		}
	} else {
		versionedcfg, err = loadConfigFromFile(*configFile)
		if err != nil {
			return nil, xerrors.Errorf("load scheduler config: %w", err)
		}
	}
	return versionedcfg, nil
}
``` 
[...]

PR #377, opened 2024-10-21
Fix Default Config Loading When ConfigFile Is Not Specified

Ianwu0812:
#### What type of PR is this?

Add related area tag(s):
/area simulator

Add one of the following kinds:
/kind bug

The `debuggableScheduler` aims to load the default configuration when no explicit `configfile` is provided. However, the current condition to check the `configFile` fails because it only verifies if the pointer is `nil`. When the pointer contains an empty string, it bypasses the default config loading logic.

To address this issue, we have added a new condition to check whether the value of `configFile` is an empty string, ensuring the default configuration is loaded as expected when no config file is specified.

Fixes https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/375

I'm preparing a presentation about kube-scheduler in Taiwan this week, I would like to use the kube-scheduler-simulator to demonstrate the mechanism. 
During preparation I found this bug, therefore I opened a PR to fix it. 
This is my first PR in CNCF, and I’m excited to contribute! If you have any questions, please feel free to reach out.

### DIFF
--- a/simulator/pkg/debuggablescheduler/debuggable_scheduler.go
+++ b/simulator/pkg/debuggablescheduler/debuggable_scheduler.go
@@ -138,7 +138,7 @@ func CreateOptionForPlugin(outOfTreePluginRegistry runtime.Registry, pluginExten
 func loadKubeSchedulerConfig(configFile *string) (*v1.KubeSchedulerConfiguration, error) {
 	var versionedcfg *v1.KubeSchedulerConfiguration
 	var err error
-	if configFile == nil {
+	if configFile == nil || *configFile == ""{
 		versionedcfg, err = simulatorschedulerconfig.DefaultSchedulerConfig()
 		if err != nil {
 			return nil, xerrors.Errorf("get default scheduler config: %w", err)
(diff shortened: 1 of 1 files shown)


sanposhiho (commented):
/ok-to-test

simulator/pkg/debuggablescheduler/debuggable_scheduler.go
```diff
@@ -138,7 +138,7 @@ func CreateOptionForPlugin(outOfTreePluginRegistry runtime.Registry, pluginExten
 func loadKubeSchedulerConfig(configFile *string) (*v1.KubeSchedulerConfiguration, error) {
 	var versionedcfg *v1.KubeSchedulerConfiguration
 	var err error
-	if configFile == nil {
+	if configFile == nil || *configFile == ""{
```
sanposhiho, line 141:
```suggestion
	if configFile == nil || *configFile == "" {
```

sanposhiho (commented):
nit

sanposhiho (approved):
/lgtm
/approve

Good luck for your presentation 😉

PR #393, opened 2024-11-12
feat: support the wasm extension in the debuggable scheduler

jgiannuzzi:
#### What type of PR is this?

/area simulator
/kind feature

This PR adds [kube-scheduler-wasm-extension](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension) support into the simulator via the debuggable scheduler. This allows easy testing of WASM plugins by simply adding the required entries to the scheduler config.

### DIFF
--- a/simulator/pkg/debuggablescheduler/debuggable_scheduler.go
+++ b/simulator/pkg/debuggablescheduler/debuggable_scheduler.go
@@ -57,6 +57,12 @@ func NewConfigs() (Configs, error) {
 		return Configs{}, xerrors.Errorf("load scheduler config: %w", err)
 	}
 
+	// Register wasm plugins to the wasm registry.
+	// This _needs_ to happen before the scheduler configuration is converted.
+	if err := simulatorschedulerconfig.RegisterWasmPlugins(versionedcfg); err != nil {
+		return Configs{}, xerrors.Errorf("register wasm plugins: %w", err)
+	}
+
 	versioned, err := scheduler.ConvertConfigurationForSimulator(versionedcfg)
 	if err != nil {
 		return Configs{}, xerrors.Errorf("convert scheduler config to apply: %w", err)
--- a/simulator/scheduler/config/wasm.go
+++ b/simulator/scheduler/config/wasm.go
@@ -1,0 +1,58 @@
+package config
+
+import (
+	"golang.org/x/xerrors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	configv1 "k8s.io/kube-scheduler/config/v1"
+	"k8s.io/kubernetes/pkg/scheduler/apis/config"
+	"k8s.io/kubernetes/pkg/scheduler/apis/config/scheme"
+	"k8s.io/kubernetes/pkg/scheduler/framework/runtime"
+	wasm "sigs.k8s.io/kube-scheduler-wasm-extension/scheduler/plugin"
+)
+
+// RegisterWasmPlugins registers wasm plugins from the given configuration.
+func RegisterWasmPlugins(versionedCfg *configv1.KubeSchedulerConfiguration) error {
+	cfg := config.KubeSchedulerConfiguration{}
+	if err := scheme.Scheme.Convert(versionedCfg, &cfg, nil); err != nil {
+		return xerrors.Errorf("convert configuration: %w", err)
+	}
+
+	registry, err := getWasmRegistryFromUnversionedConfig(&cfg)
+	if err != nil {
+		return err
+	}
+
+	SetOutOfTreeRegistries(registry)
+
+	return nil
+}
+
+// getWasmRegistryFromUnversionedConfig registers wasm plugins from the given unversioned configuration.
+func getWasmRegistryFromUnversionedConfig(cfg *config.KubeSchedulerConfiguration) (runtime.Registry, error) {
+	registry := runtime.Registry{}
+
+	for _, profile := range cfg.Profiles {
+		wasmplugins := sets.New[string]()
+		// look for the wasm plugin in the plugin config.
+		for _, config := range profile.PluginConfig {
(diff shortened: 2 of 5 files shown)


sanposhiho (commented):
/ok-to-test

simulator/scheduler/plugin/plugins.go
```diff
@@ -141,30 +141,14 @@ func NewPluginConfig(pc []configv1.PluginConfig) ([]configv1.PluginConfig, error
...
-			continue
-		}
-
-		ret = append(ret, configv1.PluginConfig{
-			Name: pluginName(name),
-			Args: *pc,
-		})
-
-		// avoid adding same plugin config.
-		delete(pluginConfig, name)
```
sanposhiho, line 167: Looks like we should keep these lines?
If there's no plugin config given from users, then we still should use `defaultpls`.

sanposhiho (commented):
/retitle feat: support the wasm extension in the debuggable scheduler

simulator/scheduler/plugin/plugins.go
jgiannuzzi: I removed these lines because they would essentially add a "wrapped" version of the plugin config, but only for the in-tree and out-of-tree plugins, as we don't know about the WASM ones at this stage yet.

By adding a "wrapped" version of the plugin config to all of them, the logic is simplified, and it ensures that wrapped WASM plugins will be able to find their config.

sanposhiho: > as we don't know about the WASM ones at this stage yet.

So.. what do you mean by this? I'm not sure how the original version didn't work out for the wasm extension plugin configuration.

simulator/pkg/debuggablescheduler/debuggable_scheduler.go
```diff
@@ -125,6 +125,14 @@ func CreateOptionForPlugin(outOfTreePluginRegistry runtime.Registry, pluginExten
 		simulatorschedulerconfig.SetOutOfTreeRegistries(outOfTreePluginRegistry)
 	}
 
+	// add wasm plugins to the out of tree registry.
+	wasmRegistry := plugin.GetWasmRegistry(internalCfg)
+	outOfTreePluginRegistry = simulatorschedulerconfig.OutOfTreeRegistries()
+	if err := outOfTreePluginRegistry.Merge(wasmRegistry); err != nil {
+		return nil, xerrors.Errorf("merge wasm registry: %w", err)
+	}
+	simulatorschedulerconfig.SetOutOfTreeRegistries(outOfTreePluginRegistry)
```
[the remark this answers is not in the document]
jgiannuzzi: that would give us only the wasm plugins, and not the ones registered by the user in https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/c5d91e0d674c721170f6ae08b1c7ef1ba7d18d14/simulator/scheduler/config/plugin.go#L10-L12

simulator/scheduler/plugin/plugins.go
jgiannuzzi: `NewPluginConfig` is called by `ConvertConfigurationForSimulator` which happens before we have a `k8s.io/kubernetes/pkg/scheduler/apis/config.KubeSchedulerConfiguration`, and we need it to be able to know which plugins are WASM plugins.

I could change the place where we add the WASM plugins to the registry to be in `NewConfigs` before the call to `ConvertConfigurationForSimulator`. That would remove the need for the changes above (which didn't sit well with me in the first place). However I'm not sure this is conceptually the right place for this either!

jgiannuzzi: Alright, I have reworked it to be able to register the wasm plugins during configuration read (`debuggablescheduler.NewConfigs`). That simplifies the implementation and decouples create the wasm plugin registry from using it.

simulator/scheduler/config/wasm.go
```diff
@@ -0,0 +1,71 @@
...
+	"golang.org/x/xerrors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	configv1 "k8s.io/kube-scheduler/config/v1"
+	"k8s.io/kubernetes/pkg/scheduler/apis/config"
+	"k8s.io/kubernetes/pkg/scheduler/apis/config/scheme"
+	"k8s.io/kubernetes/pkg/scheduler/framework/runtime"
+	wasm "sigs.k8s.io/kube-scheduler-wasm-extension/scheduler/plugin"
+)
+
+var wasmRegistry runtime.Registry
```
sanposhiho, line 14: I don't see the need of having a new variable to store the registries for wasm plugins.
Like what you did previously, can we just use the existing `outOfTreeRegistries`?

I replied the last your comment.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/393#discussion_r1838653123

simulator/pkg/debuggablescheduler/debuggable_scheduler.go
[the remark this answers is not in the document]
sanposhiho: `SetOutOfTreeRegistries` keeps the existing entries in `outOfTreeRegistries`, right?
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/70484f171910344a5c08158ee5c8a53dbfc1faae/simulator/scheduler/config/plugin.go#L57-L61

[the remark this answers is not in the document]
jgiannuzzi: Ah right, I didn't see that! Thanks for pointing that out!
Will update the code accordingly

jgiannuzzi:
Thanks for your guidance @sanposhiho. The code is now yet even simpler and feels more right 😄

sanposhiho (approved):
/lgtm
/approve

Thanks!

Issue #395, opened 2024-11-25
simulator: record events on prod cluster and replay them on a fake cluster any time


saza-ku:
/kind feature

<hr>

This issue proposes a new feature to record events on a prod cluster and replay them on a fake cluster any time.

## Background

Debugging customized schedulers is a complex challenge. One of the reasons is bugs that only occur on the prod cluster. It is hard to reproduce the issue on a fake cluster because the fake cluster does not have the same load as the prod cluster.

We have the syncer feature that makes it easy to simulate a real load on a fake cluster. However, it would be more helpful to save a series of events on the prod cluster that cause the issue and replay them on a fake cluster any time especially for debugging.

## Goals

Users can run a process that watches events on the prod cluster and saves them in some way (e.g. a JSON file). Then they can run the simulator and replay the events on a fake cluster.

### Story 1

An organization has implemented their own scheduler plugins. The plugins cause an issue only on the prod cluster. They want to reproduce the issue on a fake cluster.

#### Solution

They can record the events that cause the issue on the prod cluster. Then they can replay the events on a fake cluster to reproduce the issue.

### Story 2

They have implemented a new plugin. They want to test and evaluate it with a real load before deploying it to the prod cluster.

#### Solution

They can record the events on the prod cluster and save them. When they implement a new plugin, they can use the recorded events to test and evaluate the plugin.

## Note

This might be a fairly large feature, so please let me know if we need a KEP.

sanposhiho:
+1, thanks for a proposal!

> This might be a fairly large feature, so please let me know if we need a KEP.

No need. Please just go ahead :)

saza-ku:
Thanks:)

When replaying the events, the logic of applying resources will be the same as that of syncer. So I gonna make the logic reusable by making a new package.

But oneshotimporter also has the codes of applying resources. So how about modularizing the codes in oneshotimporter and syncer before implementing this feature?

sanposhiho:
sg, I think the applying and stuff can be in a certain package, and oneshot/syncer/replayer can use it.

saza-ku:
Okay, I'll do it first and make a PR.

saza-ku:
https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/376 is going to change `oneshotimporter`, so I first separated `resourceapplier` from `syncer`.

Next I'll implement the replaying feature using it before fixing `oneshotimporter`.

PR #400, opened 2024-12-16
Separate the logic of applying resources from syncer and make it reusable

saza-ku:
#### What type of PR is this?

/area simulator
/kind cleanup

`syncer` and `oneshotimporter` have their own logic of applying resources. They can be generalized and the replaying feature (#395) can use it.

This PR separates that of `syncer` and makes it a module `resourceapplier`.

Part of #395 

I'll create another PR to refactor `oneshotimporter` to use `resourceapplier`.

### DIFF
--- a/simulator/cmd/simulator/simulator.go
+++ b/simulator/cmd/simulator/simulator.go
@@ -23,7 +23,6 @@ import (
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/resourceapplier"
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/server"
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/server/di"
-	"sigs.k8s.io/kube-scheduler-simulator/simulator/syncer"
 )
 
 const (
@@ -95,7 +94,7 @@ func startSimulator() error {
 		return xerrors.Errorf("kubeapi-server is not ready: %w", err)
 	}
 
-	dic, err := di.NewDIContainer(client, dynamicClient, restMapper, etcdclient, restCfg, cfg.InitialSchedulerCfg, cfg.ExternalImportEnabled, cfg.ResourceSyncEnabled, importClusterResourceClient, importClusterDynamicClient, cfg.Port, syncer.Options{}, resourceapplier.Options{})
+	dic, err := di.NewDIContainer(client, dynamicClient, restMapper, etcdclient, restCfg, cfg.InitialSchedulerCfg, cfg.ExternalImportEnabled, cfg.ResourceSyncEnabled, importClusterResourceClient, importClusterDynamicClient, cfg.Port, resourceapplier.Options{})
 	if err != nil {
 		return xerrors.Errorf("create di container: %w", err)
 	}
--- a/simulator/cmd/simulator/simulator.go
+++ b/simulator/cmd/simulator/simulator.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/config"
+	"sigs.k8s.io/kube-scheduler-simulator/simulator/resourceapplier"
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/server"
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/server/di"
 	"sigs.k8s.io/kube-scheduler-simulator/simulator/syncer"
@@ -94,7 +95,7 @@ func startSimulator() error {
 		return xerrors.Errorf("kubeapi-server is not ready: %w", err)
 	}
 
-	dic, err := di.NewDIContainer(client, dynamicClient, restMapper, etcdclient, restCfg, cfg.InitialSchedulerCfg, cfg.ExternalImportEnabled, cfg.ResourceSyncEnabled, importClusterResourceClient, importClusterDynamicClient, cfg.Port, syncer.Options{})
+	dic, err := di.NewDIContainer(client, dynamicClient, restMapper, etcdclient, restCfg, cfg.InitialSchedulerCfg, cfg.ExternalImportEnabled, cfg.ResourceSyncEnabled, importClusterResourceClient, importClusterDynamicClient, cfg.Port, syncer.Options{}, resourceapplier.Options{})
 	if err != nil {
 		return xerrors.Errorf("create di container: %w", err)
 	}
--- a/simulator/resourceapplier/resourceapplier_test.go
+++ b/simulator/resourceapplier/resourceapplier_test.go
@@ -23,14 +23,16 @@ func TestResourceApplier_createPods(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name          string
-		podToApply    *corev1.Pod
-		podAfterApply *corev1.Pod
-		wantErr       bool
+		name           string
+		podToCreate    *corev1.Pod
+		podAfterCreate *corev1.Pod
+		filter         FilteringFunction
+		filtered       bool
+		wantErr        bool
 	}{
 		{
 			name: "create a Pod",
-			podToApply: &corev1.Pod{
+			podToCreate: &corev1.Pod{
 				TypeMeta: metav1.TypeMeta{
 					Kind:       "Pod",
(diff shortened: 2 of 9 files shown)


simulator/resourceapplier/resourceapplier_test.go
```diff
@@ -0,0 +1,521 @@
...
+		},
+		{
+			Group: metav1.APIGroup{
+				Versions: []metav1.GroupVersionForDiscovery{
+					{Version: "v1"},
+				},
+			},
+			VersionedResources: map[string][]metav1.APIResource{
+				"v1": {
+					{Name: "nodes", Namespaced: true, Kind: "Node"},
```
ordovicia, line 514:
```suggestion
					{Name: "nodes", Namespaced: false, Kind: "Node"},
```

Nodes are actually cluster-scope.
Having both cluster-scope and namespaced resources in the test would improve completeness.

simulator/resourceapplier/resourceapplier_test.go
```diff
@@ -0,0 +1,521 @@
...
+				t.Fatalf("failed to convert pod to unstructured: %v", err)
+			}
+			unstructedPod := &unstructured.Unstructured{Object: p}
+			err = service.Create(context.Background(), unstructedPod)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("createPods() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			got, err := client.Resource(corev1.Resource("pods").WithVersion("v1")).Namespace("default").Get(context.Background(), tt.podToApply.Name, metav1.GetOptions{})
```
ordovicia, line 92: I guess "pods", "v1", "default" should not be hard-coded, should be read from the test case instead.
Ditto in other tests.

simulator/resourceapplier/resourceapplier_test.go
```diff
@@ -0,0 +1,521 @@
...
+				return
+			}
+
+			got, err := client.Resource(corev1.Resource("pods").WithVersion("v1")).Namespace("default").Get(context.Background(), tt.podToCreate.Name, metav1.GetOptions{})
+			if err != nil {
+				if tt.filtered && errors.IsNotFound(err) || tt.wantErr {
+					return
+				}
+				t.Fatalf("failed to get pod when comparing: %v", err)
+			}
```
ordovicia, line 212: Seems that a branch should be added for `tt.filtered == true` but `err == nil` case.

saza-ku: True. And also for `err == nil && tt.wantErr`.

simulator/resourceapplier/resourceapplier_test.go
saza-ku: I added `createNodes` for the case of cluster-scope resources.

ordovicia (approved):
LGTM. Thank you!

simulator/resourceapplier/resourceapplier_test.go
```diff
@@ -0,0 +1,614 @@
...
+
+			if diff := cmp.Diff(*tt.podAfterApply, gotPod); diff != "" {
+				t.Errorf("createPods() mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestResourceApplier_createPodsWithFilter(t *testing.T) {
```
utam0k, line 109: How about merging this test function with `TestResourceApplier_createPods`? If `filter` is nil, we don't pass the Option.

saza-ku: Fixed!

utam0k:
/lgtm

@sanposhiho Would you mind checking this PR because this PR needs `/approve`

sanposhiho (commented):
This PR doesn't change oneshotimporter, did you mean you're gonna create another PR to refactor that to use resourceapplier?

simulator/cmd/simulator/simulator.go
```diff
@@ -94,7 +95,7 @@ func startSimulator() error {
 		return xerrors.Errorf("kubeapi-server is not ready: %w", err)
 	}
 
-	dic, err := di.NewDIContainer(client, dynamicClient, restMapper, etcdclient, restCfg, cfg.InitialSchedulerCfg, cfg.ExternalImportEnabled, cfg.ResourceSyncEnabled, importClusterResourceClient, importClusterDynamicClient, cfg.Port, syncer.Options{})
+	dic, err := di.NewDIContainer(client, dynamicClient, restMapper, etcdclient, restCfg, cfg.InitialSchedulerCfg, cfg.ExternalImportEnabled, cfg.ResourceSyncEnabled, importClusterResourceClient, importClusterDynamicClient, cfg.Port, syncer.Options{}, resourceapplier.Options{})
```
sanposhiho, line 98: Why do we need `resourceapplier.Options{}`? 
If you meant to use it for oneshot importer, then that's fine.
But, if you create it just in case someone wants, then we don't need it until "someone" comes to us and makes a feature request.

saza-ku: In the current implementation, `syncer` filters and mutates resources to apply and the filters and mutations can be configurable with the option. This PR moves these logic to `resourceapplier` so we need `resourceapplier`'s option to configure them.

Another possible way is callers of `resourceapplier` (`syncer`, `oneshotimporter`, `replayer` or etc) filter and mutate resources and it just applies them. But all the callers would need filters and mutations so I think it's better to commonize them as `resourceapplier`.

saza-ku:
@sanposhiho 

> This PR doesn't change oneshotimporter, did you mean you're gonna create another PR to refactor that to use resourceapplier?

Yes. This is because `oneshotimporter` was under revision when I created this PR. (https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/395#issuecomment-2544775110)

simulator/cmd/simulator/simulator.go
sanposhiho: Ah, okay I forgot about the existing options. Then you need to tweak this doc.
https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/simulator/docs/import-cluster-resources.md?plain=1

simulator/syncer/syncer.go
```diff
@@ -5,72 +5,61 @@ import (
...
+type Service struct {
+	gvrs                   []schema.GroupVersionResource
+	srcDynamicClient       dynamic.Interface
+	resourceApplierService *resourceapplier.Service
 }
 
 type Options struct {
 	// GVRsToSync is a list of GroupVersionResource that will be synced.
 	// If GVRsToSync is nil, defaultGVRs are used.
 	GVRsToSync []schema.GroupVersionResource
```
sanposhiho, line 42: Can we somehow move it to the resourceapplier option as well so that we can unify the interface of the option regardless of whether users want to use oneshot or syncer?

saza-ku: We could just make `resourceapplier` have `GVRsToSync` and `syncer` refer to `resouceApplierService.GRVsToSync`.

```go
	// in syncer service
	infFact := dynamicinformer.NewFilteredDynamicSharedInformerFactory(s.srcDynamicClient, 0, metav1.NamespaceAll, nil)
	for _, gvr := range s.resourceApplierService.GVRsToSync { // syncer itself doesn't have gvrs
		inf := infFact.ForResource(gvr).Informer()
```

`oneshotimporter` would just ignore it.

sanposhiho: > oneshotimporter would just ignore it.

Well, can't we make oneshotimporter also configurable about which resource to import with the same option of resourceapplyer?

saza-ku: Yes. So we could make a PR that makes it configurable after this PR.

For now, I'll just implement as I wrote in the last comment.

sanposhiho: Ah, got what you meant. Yes, I'm fine with it if oneshotimporter would follow the option later.

saza-ku: I unified the syncer options into the resourceapplier options.

saza-ku: And also update the doc.

sanposhiho (approved):
thanks!

/lgtm
/approve

PR #403, opened 2025-01-21
feat: implement recorder and replayer

### LINKED ISSUE #395: simulator: record events on prod cluster and replay them on a fake cluster any time

saza-ku:
#### What type of PR is this?

/area simulator
/kind feature

This PR implements recorder and replayer.

recorder continuously saves events in an external cluster into a JSON file. replayer reproduces these events in the simulator cluster when the simulator starts by reading the JSON file.

Fixes #395 


simulator/docs/record-and-replay-cluster-events.md
```diff
@@ -0,0 +1,95 @@
...
+
+You can record events from your real cluster and replay them in the simulator. This feature is useful for reproducing issues that occur in your real cluster.
+
+## Record events
+
+To record events from your real cluster, you need to follow these steps:
+
+1. Set `true` to `recorderEnabled`.
+2. Set the path of the kubeconfig file for your cluster to `KubeConfig`.
+   - This feature only requires the read permission for events.
```
sanposhiho, line 11: Does it? Looks like recorder.go have an event handers though. I guess the read permissions for all resources to be recorded are required.

simulator/docs/record-and-replay-cluster-events.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+```yaml
+volumes:
+  ...
+  - ./path/to/recorded-events.json:/path/to/recorded-events.json
+```
+
+> [!NOTE]
+> When a file already exists at `recordedFilePath`, it backs up the file in the same directory adding a timestamp to the filename and creates a new file for recording.
```
sanposhiho, line 28: Looks confusing. I prefer either just override the file without the error or make the recorder put out the error.

simulator/docs/record-and-replay-cluster-events.md
```diff
@@ -0,0 +1,95 @@
+# [Beta] Record your real cluster's events and replay them in the simulator
+
+You can record events from your real cluster and replay them in the simulator. This feature is useful for reproducing issues that occur in your real cluster.
```
sanposhiho, line 3: Please rephrase "events" in this doc. It's unclear what it means for first-readers.

simulator/docs/record-and-replay-cluster-events.md
```diff
@@ -0,0 +1,95 @@
+# [Beta] Record your real cluster's events and replay them in the simulator
```
sanposhiho, line 1: You have to put the link for this page at somewhere (README?) so that people can notice

sanposhiho (commented):
Just took a look at the doc. 
I'll take a look for the implementation after @utam0k / @ordovicia.

simulator/docs/record-and-replay-cluster-events.md
```diff
@@ -0,0 +1,95 @@
...
+
+You can record events from your real cluster and replay them in the simulator. This feature is useful for reproducing issues that occur in your real cluster.
+
+## Record events
```
sanposhiho, line 5: So, basically it has two steps, recording resources into JSON file and replaying it on the simulator.
And, people have to record resources by starting the simulator with `recorderEnabled`, then shutdown the simulator once, and then restart the simulator with the JSON file and `replayerEnabled`.

The first step looks weird for me because it has to run the simulator, but essentially it just imports the resources from a real cluster and outputs the JSON file; the simulator doesn't come into play.

So... well, I guess, should we create a CLI for it? 
It can be `recorder` (simple single cli) or `sched-simulator record` (`sched-simulator` CLI with `record` subcommand). 
Then the flow would be:
1. Users record the resource creation/update/deletion by `sched-simulator record --duration 5m` or something.
2. After 5min, CLI generates JSON file.
3. Users start the simulator with JSON file and `replayerEnabled`.

simulator/docs/record-and-replay-cluster-events.md
saza-ku: > the read permissions for all resources to be recorded are required.

That's right.
"This feature only requires the read permission for resources." was correct.

simulator/docs/record-and-replay-cluster-events.md
saza-ku: I'll fix it to put out the error.

utam0k:
> Just took a look at the doc.
I'll take a look for the implementation after @utam0k / @ordovicia.

Sure! I'll review it.

simulator/docs/record-and-replay-cluster-events.md
saza-ku: Rephrased it to "changes" or "changes in resources". How about it?

simulator/recorder/recorder.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	s.recordCh <- r
+}
+
+func (s *Service) record(ctx context.Context) {
+	records := []Record{}
+	for {
+		select {
+		case r := <-s.recordCh:
```
ordovicia, line 118: If I understand correctly, on every event (resource creation/update/delete) this method serializes all events that have happened so far and writes them to a file.
`recordCh` is unbuffered, so `RecordEvent` will be blocked until `record` completes handling an inflight event.

I guess we need some optimization like the following, otherwise the recorder would have a limited performance.

- Asynchronous: Change `recordCh` to a buffered channel so that `RecordEvent` will not be blocked.
- Buffering: Save some records in a buffer and write them together as a chunk to reduce file system operations
- Incremental: Write each record chunk into separate files to avoid writing all records many times.

simulator/docs/record-and-replay-cluster-events.md
sanposhiho: Probably go more straightforward like "You can record resource addition/update/deletion at your real cluster"

saza-ku:
@utam0k I have quite a few points to revise, so it's okay to review this PR next week :)

Issue #426, opened 2025-06-09
[bug]：ResisterResultSavingToInformer err


LY-today:
The Pod is deleted in the real cluster. When the informer detects that the update is triggered, it cannot find the Pod and reports an error.

`monitoring transient-probe-ds-4m9k9 pods "transient-probe-ds-4m9k9" not found pod`

LY-today:
@sanposhiho please check

sanposhiho:
cc @utam0k @ordovicia

LY-today:
@sanposhiho https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/428 please check

PR #428, opened 2025-06-12
fix: ResisterResultSavingToInformer add FilterFunc

LY-today:
https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/426#issuecomment-2963684307

### DIFF
--- a/simulator/scheduler/storereflector/storereflector.go
+++ b/simulator/scheduler/storereflector/storereflector.go
@@ -57,11 +57,22 @@ func (s *reflector) ResisterResultSavingToInformer(client clientset.Interface, s
 	informerFactory := scheduler.NewInformerFactory(client, 0)
 	// Reflector adds scheduling results when pod is updating.
 	// This is because Extenders doesn't have any phase to hook scheduling finished. (both successfully and non-successfully)
-	_, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(
-		cache.ResourceEventHandlerFuncs{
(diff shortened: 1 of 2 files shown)


simulator/scheduler/storereflector/storereflector.go
```diff
@@ -57,11 +57,22 @@ func (s *reflector) ResisterResultSavingToInformer(client clientset.Interface, s
 	informerFactory := scheduler.NewInformerFactory(client, 0)
 	// Reflector adds scheduling results when pod is updating.
 	// This is because Extenders doesn't have any phase to hook scheduling finished. (both successfully and non-successfully)
-	_, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(
-		cache.ResourceEventHandlerFuncs{
+	_, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: func(obj interface{}) bool {
+			if _, ok := obj.(cache.DeletedFinalStateUnknown); ok {
```
utam0k, line 62:
```suggestion
            if obj == nil {
                klog.V(5).Info("Filtering nil object")
                return false
            }

			if _, ok := obj.(cache.DeletedFinalStateUnknown); ok {
```

LY-today: Should we directly remove the judgment of cache.DeletedFinalStateUnknown?

LY-today:
> Can't we just cut out this logic and test it?

Which part of the logic do you mean to delete and then test?

LY-today:
@utam0k Added unit tests

Issue #421, opened 2025-06-09
[bug]：serviceaccount not found


LY-today:
When connecting to a real cluster, the following error is reported. Do you need to add a serviceaccount synchronization mechanism?

```
failed to apply Pod: pods "cert-manager-cainjector-f9648d4cf-9nt47" is forbidden: error looking up service account cert-manager/cert-manager-cainjector: serviceaccount "cert-manager-cainjector" not found
```

LY-today:
@sanposhiho please check

sanposhiho:
Thanks for all the bug reports, by the way! @LY-today The syncing feature is still a baby, and we need to refine it.

LY-today:
> For a short term solution, can you just try to use [`resourceApplierOptions.GVRsToApply`](https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/simulator/docs/import-cluster-resources.md) so that the simulator syncs SA as well?
> 
> But, either way, we need to solve this problem because using SA with pods is very common in real clusters. The possible options are:
> 
> * syncing SA additionally by default
> * not syncing SA additionally by default and clearing SA in pod spec. Only when `resourceApplierOptions.GVRsToApply` has SA, we don't clear SA.
> 
> The former is simpler, but would increase the number of objects that we have to sync even though SA isn't necessary for the default scheduling. So, those are pros and cons.

Agree, there are indeed pros and cons, so you can evaluate the best solution. I can provide MR

sanposhiho:
@saza-ku @utam0k @ordovicia What is your take? I'm inclined to the first option, just syncing SA additionally because I feel like the latter makes the implementation complicated for a small win.

sanposhiho:
> We've implemented so that it removes service accout name from pod specs, haven't we?

Hmm, why is this problem happening then? 
Anyways, I think we can just start syncing SA as well. @LY-today can you open a PR, then we can review.

LY-today:
> > We've implemented so that it removes service accout name from pod specs, haven't we?
> 
> Hmm, why is this problem happening then? Anyways, I think we can just start syncing SA as well. [@LY-today](https://github.com/LY-today) can you open a PR, then we can review.

@sanposhiho I was thinking of doing that, but can I submit my PR tomorrow? Today is not convenient.

sanposhiho:
Of course, you can do it whenever you're convenient. Take your time.

Li-Qri:
[...]

PR #432, opened 2025-06-16
fix: add applySas

LY-today:
https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/421

### DIFF
--- a/simulator/oneshotimporter/importer.go
+++ b/simulator/oneshotimporter/importer.go
@@ -31,6 +31,7 @@ type Service struct {
 var DefaultGVRs = []schema.GroupVersionResource{
 	{Group: "", Version: "v1", Resource: "namespaces"},
 	{Group: "scheduling.k8s.io", Version: "v1", Resource: "priorityclasses"},
+	{Group: "", Version: "v1", Resource: "serviceaccounts"},
 	{Group: "storage.k8s.io", Version: "v1", Resource: "storageclasses"},
 	{Group: "", Version: "v1", Resource: "persistentvolumeclaims"},
 	{Group: "", Version: "v1", Resource: "nodes"},
--- a/simulator/snapshot/snapshot.go
+++ b/simulator/snapshot/snapshot.go
@@ -5,6 +5,7 @@ package snapshot
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 
 	"golang.org/x/xerrors"
@@ -42,6 +43,7 @@ type ResourcesForSnap struct {
 
 // ResourcesForLoad indicates all resources and scheduler configuration to be loaded.
 type ResourcesForLoad struct {
+	Sas             []v1.ServiceAccountApplyConfiguration             `json:"sas"`
 	Pods            []v1.PodApplyConfiguration                        `json:"pods"`
 	Nodes           []v1.NodeApplyConfiguration                       `json:"nodes"`
 	Pvs             []v1.PersistentVolumeApplyConfiguration           `json:"pvs"`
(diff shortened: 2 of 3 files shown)


sanposhiho (commented):
/ok-to-test
/cc @saza-ku @ordovicia @utam0k

simulator/oneshotimporter/importer.go
```diff
@@ -31,6 +31,7 @@ type Service struct {
 var DefaultGVRs = []schema.GroupVersionResource{
 	{Group: "", Version: "v1", Resource: "namespaces"},
 	{Group: "scheduling.k8s.io", Version: "v1", Resource: "priorityclasses"},
+	{Group: "", Version: "v1", Resource: "serviceaccounts"},
```
saza-ku, line 34: Could you add it also to syncer's DefaultGVRs?

simulator/snapshot/snapshot.go
```diff
@@ -42,6 +42,7 @@ type ResourcesForSnap struct {
 
 // ResourcesForLoad indicates all resources and scheduler configuration to be loaded.
 type ResourcesForLoad struct {
+	Sas             []v1.ServiceAccountApplyConfiguration             `json:"sas"`
```
saza-ku, line 45: Now `SnapshotService` is only used by `SnapshotHandler` and it only handles resources of the fake cluster. So we don't need to change `SnapshotService`.

saza-ku (commented):
Could you add test cases of importing/syncing pods with account services?

simulator/oneshotimporter/importer.go
LY-today: done

simulator/snapshot/snapshot.go
LY-today: Do I need to delete this line?

LY-today: There will be an exception after deletion

image

LY-today:
> Could you add test cases of importing/syncing pods with account services?

It seems that the existing single test of simulator/snapshot cannot run smoothly? It seems that there is a go.mod dependency conflict?

simulator/snapshot/snapshot.go
saza-ku: I mean we don't need the entire fix in `SnapshotService`, not only the line.

simulator/oneshotimporter/importer.go
saza-ku: You may forget to push it. `syncer` is not fixed now.

saza-ku:
Does this error occur when you run `make test`? I did it on my machine, but it didn't repro.

simulator/snapshot/snapshot.go
LY-today: > I mean we don't need the entire fix in `SnapshotService`, not only the line.

Wow, I seem to be losing my mind, I don't understand your question at all

saza-ku: First, #421 occurs when you use [oneshotimporter and syncer](https://github.com/kubernetes-sigs/kube-scheduler-simulator/blob/master/simulator/docs/import-cluster-resources.md), doesn't it? And these features don't depend on `SnapshotService`. So in my understanding, #421 isn't related to `SnapshotService`.
If it's wrong, please tell me why we need to edit `SnapshotService`.

LY-today:
> Does this error occur when you run `make test`? I did it on my machine, but it didn't repro.

What is your go version?

simulator/snapshot/snapshot.go
Li-Qri, line 45: I also think Explicitly setting the ServiceAccount to default is sufficient. Synchronizing ServiceAccounts might introduce additional issues.

LY-today: This has been discussed in issues before. After everyone reaches a consensus, synchronization is required.

LY-today: > Almost the same as `Load`. You would add `Sas` to `ResourcesForSnap`, and implement `listSas`.

done

LY-today:
@saza-ku ci reports this error, how can I adjust it?
snapshot/snapshot.go:105:1: calculated cyclomatic complexity for function get is 11, max is 10 (cyclop)

PR #433, opened 2025-06-16
fix: add DefaultGVRs

LY-today:
https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/421


sanposhiho:
Can you just include this change into #432.

LY-today:
> Can you just include this change into #432.

done

PR #435, opened 2025-06-16
fix: add FilterFunc to skip deleted Pods in ResisterResultSavingToInf…

### LINKED ISSUE #426: [bug]：ResisterResultSavingToInformer err

Park-Jiyeonn:
### What this PR does

This PR adds a `FilterFunc` to the `ResisterResultSavingToInformer` method to prevent scheduling results from being written to Pods that are being deleted.

### Why

As described in [issue #426](https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/426#issuecomment-2963684307), when a Pod is being deleted from the cluster, the informer may still receive update events, but attempts to annotate such Pods will fail because they no longer exist.

By adding a `FilterFunc`, we ensure that only Pods not marked for deletion (`DeletionTimestamp == nil`) are processed, preventing errors during update handling.

### Related Issue

Fixes: [#426](https://github.com/kubernetes-sigs/kube-scheduler-simulator/issues/426)

### Additional Notes

- This change prevents unnecessary update attempts on terminating Pods.
- Ensures the simulator behaves more robustly in real cluster environments.

### DIFF
--- a/simulator/scheduler/storereflector/storereflector.go
+++ b/simulator/scheduler/storereflector/storereflector.go
@@ -57,11 +57,19 @@ func (s *reflector) ResisterResultSavingToInformer(client clientset.Interface, s
 	informerFactory := scheduler.NewInformerFactory(client, 0)
 	// Reflector adds scheduling results when pod is updating.
 	// This is because Extenders doesn't have any phase to hook scheduling finished. (both successfully and non-successfully)
-	_, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(
-		cache.ResourceEventHandlerFuncs{
+	_, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: func(obj interface{}) bool {
+			if pod, ok := obj.(*corev1.Pod); ok {
+				if pod.DeletionTimestamp != nil {
+					return false
+				}
+			}
+			return true
+		},
+		Handler: cache.ResourceEventHandlerFuncs{
 			UpdateFunc: s.storeAllResultToPodFunc(client),
 		},
-	)
+	})
 	if err != nil {
 		return xerrors.Errorf("failed to AddEventHandler of Informer: %w", err)
 	}
@@ -75,7 +83,7 @@ func (s *reflector) ResisterResultSavingToInformer(client clientset.Interface, s
 // storeAllResultToPodFunc returns the function that reflects all results on the pod annotation when the scheduling is finished.
 // It will be used as the even handler of resource updating.
 //
-//nolint:gocognit,cyclop
+//nolint:gocognit,cyclop,funlen
 func (s *reflector) storeAllResultToPodFunc(client clientset.Interface) func(interface{}, interface{}) {
 	return func(_, newObj interface{}) {
 		ctx := context.Background()
@@ -91,6 +99,9 @@ func (s *reflector) storeAllResultToPodFunc(client clientset.Interface) func(int
 			// the shared informer.
 			newPod, err := client.CoreV1().Pods(pod.Namespace).Get(ctx, pod.Name, metav1.GetOptions{})
 			if err != nil {
+				if apierrors.IsNotFound(err) {
+					return true, nil
+				}
 				return false, xerrors.Errorf("get pod: %w", err)
 			}
 			if newPod.UID != pod.UID {
@@ -124,6 +135,9 @@ func (s *reflector) storeAllResultToPodFunc(client clientset.Interface) func(int
 
 			_, err = client.CoreV1().Pods(pod.Namespace).Update(ctx, pod, metav1.UpdateOptions{})
 			if err != nil {
+				if apierrors.IsNotFound(err) {
+					return true, nil
+				}
 				// Even though we fetched the latest Pod object, we still might get a conflict
 				// because of a concurrent update. Retrying these conflict errors will usually help
 				// as long as we re-fetch the latest Pod object each time.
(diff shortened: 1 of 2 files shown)


sanposhiho (commented):
/cc @utam0k @ordovicia @saza-ku 
/ok-to-test

saza-ku (commented):
Thank you for your contribution.

While filtering out Pods where `DeletionTimestamp != nil` can mitigate the issue, it doesn't prevent it completely. A Pod could still be deleted during the execution of UpdateFunc even if its `DeletionTimestamp` was `nil` at the beginning. This suggests that it might be impossible to avoid this race condition entirely.

As an alternative, what do you think about configuring the `FilterFunc` so that the `UpdateFunc` is executed only when necessary? The `UpdateFunc` should run when a Pod has been successfully scheduled. Therefore, I suggest triggering the `UpdateFunc` only when the `nodeName` of the Pod has changed between the update, as this indicates that scheduling is complete.

saza-ku:
Ah, you're right. This will prevent from recording filtered out scheduling results.

So I wonder if this fix eliminates the error reported by #426.  
@LY-today Do you think this resolves your issue?

LY-today:
> Ah, you're right. This will prevent from recording filtered out scheduling results.
> 
> So I wonder if your fix eliminates the error reported by #426. @LY-today Do you think this resolves your issue?


can be solved

Park-Jiyeonn:
Hi @saza-ku 

Just wanted to kindly follow up on this PR 🙇  
Let me know if there's anything else I should revise or clarify.

Thank you again for your time and feedback!

Park-Jiyeonn:
@sanposhiho 

Hope you're doing well! Just wanted to follow up on this PR again. I'd really appreciate it if you could take a look or let me know if there's anything else I should improve 🙇

Looking forward to your feedback. Thank you so much for your time!

Park-Jiyeonn:
@sanposhiho @saza-ku PTAL

Park-Jiyeonn:
@sanposhiho @saza-ku PTAL, Thx!

sanposhiho:
Sorry, I noticed you pinged me many times, but I got my laptop broken and hence will be away from oss work until I get a new one (will arrive this week or next).

saza-ku:
@Park-Jiyeonn Sorry for being late.

Actually I'm uncertain about the purpose of this change because of the following reasons.

- This PR does not fully eliminate the error (as I mentioned at https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/435#pullrequestreview-2948189003)
- In my understanding even with the error present, the simulator operates without any noticeable issues.

Could you please provide more context or elaborate on the reasoning behind this change?

Park-Jiyeonn:
> @Park-Jiyeonn Sorry for being late.
> 
> Actually I'm uncertain about the purpose of this change because of the following reasons.
> 
> * This PR does not fully eliminate the error (as I mentioned at [fix: add FilterFunc to skip deleted Pods in ResisterResultSavingToInf… #435 (review)](https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/435#pullrequestreview-2948189003))
> * In my understanding even with the error present, the simulator operates without any noticeable issues.
> 
> Could you please provide more context or elaborate on the reasoning behind this change?

Regarding the concern you mentioned — I believe I addressed this in my earlier comment [#issuecomment-2995142507](https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/435#issuecomment-2995142507), but let me briefly summarize again:

* The main goal of this change is to prevent unnecessary update attempts on Pods that are already terminating (DeletionTimestamp != nil), which caused the error in #426.
* Even though the simulator continues to operate with the error, this fix improves robustness and reduces noise in logs during scheduling simulations.

sanposhiho:
> but attempts to annotate such Pods will fail because they no longer exist.

Why not just ignore NotFound error when applying?

Park-Jiyeonn:
> > but attempts to annotate such Pods will fail because they no longer exist.
> 
> Why not just ignore NotFound error when applying?

That’s exactly the purpose of this change, I added the `FilterFunc` is to ignore those NotFound errors

sanposhiho:
But, that is not perfect depending on the timing. If we just ignore the not found errors at the fetch API, that's perfect prevention, no?

Park-Jiyeonn:
> But, that is not perfect depending on the timing. If we just ignore the not found errors at the fetch API, that's perfect prevention, no?

do you mean we don't need the `filterFunc`?

sanposhiho:
I'd say we can keep your filter to reduce the num of fetch API calls (performance benefit), and we can add an additional error handling to ignore not-found errors for a perfect prevention of the issue.

Park-Jiyeonn:
> I'd say we can keep your filter to reduce the num of fetch API calls (performance benefit), and we can add an additional error handling to ignore not-found errors for a perfect prevention of the issue.

How about add a warning before `return false`?

```go
         FilterFunc: func(obj interface{}) bool {
			if pod, ok := obj.(*corev1.Pod); ok {
				if pod.DeletionTimestamp != nil {
                    add some warning or error
					return false
				}
			}
```

sanposhiho:
I don't think we need warning. Warning implies something goes wrong, but in this case, nothing goes wrong: It's just that the pod is being deleted (i.e., users would have no action to solve this warning)

Park-Jiyeonn:
> I don't think we need warning. Warning implies something goes wrong, but in this case, nothing goes wrong: It's just that the pod is being deleted (i.e., users would have no action to solve this warning)

it's OK! Additionally ignore NotFound on Get/Update.

sanposhiho (approved):
/lgtm
/approve

